PR #60, opened 2016-09-17
Adding new native (Get5_ChangeMatchState)

PreFix:


splewis:
Changing the g_GameState gamestate (which is what this does) isn't a good idea, and won't have the effect you want it to have. For example, looking at the states:
```
enum GameState {
    GameState_None,                          // no setup has taken place
    GameState_PreVeto,                       // warmup, waiting for the veto
    GameState_Veto,                          // warmup, doing the veto
    GameState_Warmup,                        // setup done, waiting for players to ready up
    GameState_KnifeRound,                    // in the knife round
    GameState_WaitingForKnifeRoundDecision,  // waiting for a .stay/.swap command after the knife round
    GameState_GoingLive,                     // in the lo3 process
    GameState_Live,                          // the match is live
    GameState_PostGame,                      // postgame screen + waiting for GOTV to finish broadcast
};
```

If the current state is GameState_None, and you call this native with any non-None arg the game won't actually be setup and the plugin will be in a bogus state.

If the current state is GameState_PreVeto, and you call this native with something like GameState_GoingLive, the plugin won't actually start. It'll just be broken.

If the current state is GameState_Live, and you call this native with something like GameState_PostGame, the game won't end. It won't stop the demo, it won't kick players, it'll just be broken.

What, specifically, are you trying to accomplish with this native?

splewis:
Closing since this change doesn't seem good, and there's no response to my question. You can reopen if you want more discussion.

PR #65, opened 2016-09-30
Fix broken query

toomuchio:
Query was broken resulted in errors being thrown eg.
L 09/30/2016 - 20:26:56: [get5_mysqlstats.smx] Last Connect SQL Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'firstkill_t, firstkill_ct, firstdeath_t, firstdeath_ct, ) VALUES (6, 0, '**' at line 1


splewis:
Thanks for the fix.

PR #70, opened 2016-10-12
Fix pauses when pauses left is equal to 0.

Technoblazed:
Probably did it a really bad way, but gave it a go nonetheless. When max pauses is set to 0, the value of pauses used continues to go up as there is no max cvar value. This is an attempt at fixing that.


splewis:
Looks good to me.

PR #73, opened 2016-10-15
Addition of ability to use custom prefix

Technoblazed:
This looks like it was removed a while back, was there a reason for this? Still seems like something that should be in the plugin :)

EDIT: Found your post regarding this, added a powered by message.


splewis:
I didn't merge this directly since my merge conflict skills proved terrible, but this is largely added (with some changes). Sorry I wasn't able to merge directly.

Technoblazed:
np

PR #74, opened 2016-10-15
Addition of LO3, including cvar to toggle

Technoblazed:
I'm aware this was here before, but you said if someone readded with a cvar you'd readd it, so here you go :)

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -232,7 +232,7 @@ public void OnPluginStart() {
     g_MaxPauseTimeCvar = CreateConVar("get5_max_pause_time", "300",
         "Maximum number of time the game can spend paused by a team, 0=unlimited");
     g_QuickRestartCvar = CreateConVar("get5_use_lo3", "0",
-        "Whether to use a quick restart or a full live-on-3 restart.");
+        "Whether to do a live-on-3 restart in addition to the warmup countdown.");
     g_ResetPausesEachHalfCvar = CreateConVar("get5_reset_pauses_each_half", "1",
         "Whether pause limits will be reset each halftime period");
     g_PausingEnabledCvar = CreateConVar("get5_pausing_enabled", "1",
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -64,6 +64,7 @@ ConVar g_MaxBackupAgeCvar;
 ConVar g_MaxPausesCvar;
 ConVar g_MaxPauseTimeCvar;
 ConVar g_PausingEnabledCvar;
+ConVar g_QuickRestartCvar;
 ConVar g_ResetPausesEachHalfCvar;
 ConVar g_StatsPathFormatCvar;
 ConVar g_StopCommandEnabledCvar;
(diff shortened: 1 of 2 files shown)


splewis:
One of the reasons I didn't want to do this myself is that it could cause issues with other systems: like the stats system (https://github.com/splewis/get5/blob/master/scripting/get5/stats.sp#L52)

I'm not confident that this change wouldn't cause round starts to incorrectly change the roundsplayed stat.

Technoblazed:
I'll be testing it in about 30 minutes, would you like me to get back to you about it?

scripting/get5.sp
```diff
@@ -230,6 +231,8 @@ public void OnPluginStart() {
         "Maximum number of pauses a team can use, 0=unlimited");
     g_MaxPauseTimeCvar = CreateConVar("get5_max_pause_time", "300",
         "Maximum number of time the game can spend paused by a team, 0=unlimited");
+    g_QuickRestartCvar = CreateConVar("get5_quick_restarts", "0",
```
splewis: "quick restarts" isn't a good name for this. get5_use_lo3 would be much descriptive.

scripting/get5.sp
```diff
@@ -231,8 +231,8 @@ public void OnPluginStart() {
         "Maximum number of pauses a team can use, 0=unlimited");
     g_MaxPauseTimeCvar = CreateConVar("get5_max_pause_time", "300",
         "Maximum number of time the game can spend paused by a team, 0=unlimited");
-    g_QuickRestartCvar = CreateConVar("get5_quick_restarts", "0",
-        "Whether to use a quick restart or a full live-on-3 restart. 0 for LO3.");
+    g_QuickRestartCvar = CreateConVar("get5_use_lo3", "0",
+        "Whether to use a quick restart or a full live-on-3 restart.");
```
splewis: I don't think a "quick restart" really describes what happens when this is 0. "Whether to do a live-on-3 restart in addition to the warmup countdown" would be better.

Technoblazed:
Apologies for the delay

splewis:
(don't worry about the merge conflicts I caused)

I think this is still confusing to use, it does the X (default 10) second countdown, then 2 restarts starting at "restart 2/3", which is odd behavior. A fix would be something like also printing "round 1/3" at the initial warmup countdown, though I still don't know if all this is worth the added confusion.

splewis:
Sorry I've left this hanging for so long.

I think I've decided not to add this. The main reason is that I'm trying to avoid including "invasive" things into this plugin. Using the game's builtin warmup system/countdown feels much more natural (even if pepole have used lo3 configs for many years). Another example of an "invasive" thing I don't want would be things like printing all the client money in chat - I'd like to avoid these types of things if there's a good alternative. 

The other reason would the slight awkwardness in the implementation of all the timing of the restarts (arbitrary values and lots of ``CreateTimer``'s). Those are hard to avoid, but I don't think having this option is valuable to deal with all that mess.

PR #75, opened 2016-10-15
Prevent stop command being used during half time

Technoblazed:


scripting/get5.sp
```diff
@@ -847,7 +847,10 @@ public Action Command_Stop(int client, int args) {
     }
 
     int roundsPlayed = GameRules_GetProp("m_totalRoundsPlayed");
-    if (g_GameState != GameState_Live || roundsPlayed == 0) {
+    int roundsPerHalf = GetCvarIntSafe("mp_maxrounds") / 2;
+
+    // ( Match Not Live || No Rounds Played || Half Time || Full Time, 2 * Half Time)
+    if (g_GameState != GameState_Live || roundsPlayed == 0 || roundsPlayed == roundsPerHalf || (roundsPlayed == (2 * roundsPerHalf))) {
```
splewis: I would think just checking g_PendingSideSwap would be sufficient here (in addition to the original checks).

Technoblazed: Oh yeah...

scripting/get5.sp
```diff
@@ -846,8 +846,7 @@ public Action Command_Stop(int client, int args) {
         return Plugin_Handled;
     }
 
-    int roundsPlayed = GameRules_GetProp("m_totalRoundsPlayed");
-    if (g_GameState != GameState_Live || roundsPlayed == 0) {
+    if (g_GameState != GameState_Live || g_PendingSideSwap == true ) {
```
splewis: Have you checked to see what happens if someone tries to use !stop the first round of a game? The plugin should be able to work (the game does generate a backup file for the start of round 0, as far as I know), but I'm not sure that it is supported correctly yet.

Technoblazed: I can't say I tested that with the new updated code, did it before and you couldn't do it on the first round, let me take a look.

Technoblazed: Stop worked as normal on the first round by the looks of it, just went back to  0:0

PR #120, opened 2017-03-17
Add "veto_first" setting and implementation

miped:
Adds a setting on the root level of the match config called "veto_first"
which allows for selecting the team to get first veto.

Not setting it, or setting it to anything but "team1" or "team2" will
default it to "team1", and thus work as it does currently.


splewis:
Looks good to me. Thanks!

Issue #122, opened 2017-03-18
Veto's can come in a little hot


miped:
A lot of CS players have what seems to amount to a nervous tick with constantly switching weapons (i.e. pressing  1 through 5 constantly). When running MWL, we were very clear that veto's would be happening as soon as the last player readied up, and that we would not be redoing veto's because people weren't paying attention and vetoed maps by mistake.

Quite a number of people hit that situation, and we've gotten feedback both at the event, and in the post-event survey that the veto does come in a little bit hot. While we could just take a hardline approach on this, i think there are some easy ways to make it a little more foolproof.

They are not mutually exclusive, so i think we should add both of them.

Proposal 1
----------
Do a 5 second countdown and a restart of the match (still in warmup mode), but freeze all the players while the veto is happening. Haven't looked into how easy this would be to implement, but there are several examples in the code about restarting the game, so should be fairly straight forward.

Basically, there would be messages to all players, and the game would restart which would go something like
```
[Get5] Veto coming up in 5...
[Get5] Veto coming up in 4...
[Get5] Veto coming up in 3...
[Get5] Veto coming up in 2...
[Get5] Veto coming up in 1...
<game restarts>
<match pauses during freezetime>
[Get5] Veto started. Game will unpause when veto is complete.
```
And the game will be unpaused if veto is aborted (on a captain drop). If veto is finished successfully, then the map change will happen right after, and there's no reason to unpause the game.

Proposal 2
-----------
Once one of the veto menus is presented, record a timestamp. When the captain does a selection, figure out how fast that pick was made. If they selected an option in less than 3 seconds, present another menu that goes like this:
```
Are you sure you want to ban Cache:
  8. Yes, ban Cache
  9. No, go back
```

The important thing here is that the options are on buttons which most players don't use very often. And adding the timer means that if they actually thought about it, they won't be presented with this menu.

Technoblazed:
The first half of `Option 1` would be the better option, but not the restarting of the game and pausing section of it, as warmup should still continue during the veto stages.

miped:
I don't agree, for the following reasons:
 - Veto is important, and everyone should be paying attention and communicating. Not running around warming up.
 - There will be another round of warmup once the veto is done. Players should be able to chill out for like 2 min while vetoing is happening.
[...]

PR #126, opened 2017-03-19
#122 Add configurable veto confirmation

miped:
Introduces a new setting get5_veto_confirmation_time, which is the time
in seconds from a menu has been presented to a choice is made, that will
result in a confirmation menu.

If set to 0, confirmations will never be shown. If set to a high number
(e.g. 3600 for an hour), confirmations will always be shown. Set to a
small value (e.g. 1 or 2) to prevent accidental vetos/picks, when a
player doesn't notice the menu while spamming buttons `1` through `5`.

Menu titles and options are customizable and translatable, and the
yes/no options are placed on button 8 and 9 respectively, to make sure
they won't get chosen by mistake.

Relates to: splewis/get5#122

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -72,7 +72,7 @@ ConVar g_TeamTimeToKnifeDecisionCvar;
 ConVar g_TeamTimeToStartCvar;
 ConVar g_TimeFormatCvar;
 ConVar g_VetoConfirmationTimeCvar;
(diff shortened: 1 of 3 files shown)


miped:
I've resolved the merge conflicts that happened after veto countdown got merged.

splewis:
Look good. Thanks!

Issue #118, opened 2017-03-05
Add option for waiting on spectator ready up


splewis:

miped:
Hmm, this seems to be mostly implemented, except for a config to actually turn it on?

splewis:
It still needs some extra logic to check if MatchTeam_TeamSpec is ready: see https://github.com/splewis/get5/blob/master/scripting/get5/readysystem.sp#L124

miped:
Ok, i'll take a look at it tomorrow and see if i can get it ready for use.

splewis:
Fixed by #127

Issue #119, opened 2017-03-15
BO3 or BO5 with map advantage


miped:
We'd like a way to do a 1 map advantage when doing a BO3 or a BO5. Currently, we've tried using series_score in the team settings to do this, but it interferes with veto's in a funky way. So this is a proposal for a couple of different ways to get around this.

Sorry in advance for the wall of text :-)

Current issue with using series_score for advantage:
----------------------------------------------------
Since the first map is the one being skipped, if you give the higher seeded team the first veto this is what will go down:

 - Map1 banned (team1)
 - Map2 banned (team2)
 - Map3 picked (team1) -- NOT PLAYED
 - Map4 picked (team2) -- PLAYED FIRST

Thus, there is no advantage to vetoing first, as you won't get to play your pick first, and if you don't know how what will happen, you'll be confused when your first pick isn't played.

If you give the lower seeded team the first veto, in order to let the higher seeded team play their pick first, it looks like this:

 - Map1 banned (team2)
 - Map2 banned (team1)
 - Map3 picked (team2) -- NOT PLAYED (essentially a third ban!)
 - Map4 picked (team1) -- PLAYED FIRST

So while the higher seeded team do get the benefit of playing their map choice first, the lower seeded team now gets 2 bans instead of one, and they could potentially choose a map to skip that they know the higher seeded team would prefer to play.

Issues with using series_score as a backup solution:
----------------------------------------------------
If you do want to use series_score as a backup solution you'd need to:

 - Turn off vetos (works)
 - Select exactly the map list to be played (including the maps to skip) (works)
 - Unless you use `always_knife` as your side type, you'd need to specify which side each team starts on the same way as they chose in their veto (not possible)

Since it's not possible to specify sides I would argue that it has limited use as a backup solution.

So, with that out of the way, and after looking through the code, i think there are a couple of ways to make this work better as a solution for map advantage

Proposal 1
----------
This is the simplest of the options, since it will reuse most of the current series_score functionality.

Basically, it will add this little piece of logic to the veto process (after vetoing is complete):
```
maplist = <finished map list>
total_series_score = team1_score + team2_score
for i = 0; i < total_series_score; i++:
    # get the last map in the map list to play
    last_map = maplist.pop()
    # put it in the front of the list
    maplist.shift(last_map)
```

This will make sure that the maps that will be skipped are the ones that were picked last (or the left over map). For the BO3 case, it would make sense to just skip the last 2 bans, since they won't have any impact anyway.

[...]

PR #127, opened 2017-03-19
#118 - spectator ready

miped:
So this started as "adding spectators to ready system", but ended up as a pretty big overhaul. I think it's become a little clearer and better organised in the process though.

It's implemented by adding a key to the match config called "min_spectators_ready", which is set to `0` by default. If you set it to `1` or higher then it requires that number of spectators to ready up before the match can start. The reason for not using the number of spectators given in the match config auth list, is that often times you'll want to add a couple of spectators for actually spectating + any number of admins, which might join as spectators if necessary, but won't always be connected if things are running smoothly.

Doing it this way means that it's completely backwards compatible, and there shouldn't be any visible changes unless it's enabled.

Currently, spectators are only required to ready up for the game going to knife/live, but that could easily be changed (or be made configurable), if we wanted the same functionality for vetos or backup restores.

A couple of things that I found along the way:
 - Coaches weren't being handled consistently. This pull requests makes it so coaches are always excluded from the ready system (i.e. they don't count).
 - The spectator team wasn't given a name, which made some of the different messages weird when they were folded into the ready system. They're now called `casters` by default, but it can be changed in the match config.
 - There are a couple of utility methods in different places in the plugin that do essentially the same thing, mostly related to counting and finding players. Would be nice to clean that up.
 - Related, it's not completely clear when we use `IsPlayer` vs `IsAuthedPlayer`, but since most of the existing code seemed to use `IsPlayer`, that's what now used consistently throughout the ready system.
 - There are some parts of what could be considered the ready system that are in the main `get5.sp` file. Not sure if they should be moved?

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Of the below fields, only the ``team1`` and ``team2`` fields are actually requir
 - ``favored_percentage_team1``: wrapper for ``mp_teamprediction_pct``
 - ``favored_percentage_text`` wrapper for ``mp_teamprediction_txt``
 - ``cvars``: cvars to be set during the match warmup/knife round/live state
-- ``spectators``: see the team schema below (only the ``players`` section is used for spectators)
+- ``spectators``: see the team schema below (only the ``players`` and ``name`` sections are used for spectators)
 - ``team1``: see the team schema below
 - ``team2``: see the team schema below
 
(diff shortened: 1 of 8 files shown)


Technoblazed:
Wow.

miped:
Yeah it's a bit of a big'un, but i found it pretty hard to keep the changes that were needed contained to small pieces. Ended up becoming pretty far reaching.

splewis:
Overall looks good, I'll try to give it a more thorough look-through tomorrow before I merge (same with #119).

PR #162, opened 2017-07-02
Update import_stats.sql

AlexBksiad:
07/02/2017 - 16:11:17: [get5_mysqlstats.smx] Failed match creation query, error = Field 'end_time' doesn't have a default value

Fixes this error.
You're all welcome


AlexBksiad:
Basically the error is caused because you have a NOT NULL setting on the end_time attribute, when by default the end_time attribute is designed to not be filled in until later. To resolve this we just remove NOT NULL option for those specific attributes.

splewis:
Thanks!

Issue #160, opened 2017-06-29
Readyup timer isn't synced when one team choose to ready and then becomes unready again.


grenadecx:
If making a feature request, you should delete all the pre-filled text here.
If reporting a bug, fill in the following sections:

### Expected behavior
When enabling max 5 minutes timer for the readycheck (or else forfeit) and ct becomes ready, waits 30 seconds and becomes unready. The time is suppose to be the same for ct and t. I also expect the time to be announced at the same time for the teams.

### Actual behavior
When enabling max 5 minutes timer for the readycheck (or else forfeit) and ct becomes ready, waits 30 seconds and becomes unready. The time for forfeit is now different for t and ct, where ct have a longer timer which is now incorrect. Announcment of the time left is also no longer announced at the same time for the teams.

### Steps to reproduce
Set get5_time_to_start to 300.
Start game, and as ct, ready up, wait 30 seconds and unready. The time for both teams will no longer match and they will not be announced at the same time.

- Plugin version: 0.5.3-dev
- Sourcemod version: 1.8.0.6001

splewis:
Thanks for the clear bug report!

I don't have time to look into it now, but I'll try to get around to it in around a week.

miped:
I think this is a misunderstanding of how the feature is meant to be used. It is meant to be per team. This means that it's not a "global" 5 minutes to start for instance.

The idea is that each team has a certain amount of time to ready up. Once they've used up that time, they will forfeit the match. It is done this way so that teams have an incentive to ready up. If you make the timer global, then this can happen:

* Team 1 joins and readies up immediately. They're here to play, and they just wanna get started.
* Team 2 joins and dicks around. Doesn't ready up.
* Global timer reaches zero, because team 2 is still dicking around. 
* Match is stopped without any winner, or even worse, the wrong team is forfeited.

It would be fine to add a global ready time as well, but that's a different feature from the one implemented (which works as expected btw).

splewis:
@miped 

In your example, the match should just be a forfeit for team2 since they're not ready at (match_init_time + k seconds). That seems reasonable to me, or am I missing something?

(I do agree this could be seen as working correctly and being a separate feature request, and I could imagine a cvar for picking between whether ready up time is global or per-team, but this global method seems correct to me)

miped:
Well you could argue that, but leaving it global also gives a team the option of "running down the clock".

Consider this:
* Global timer is 5 minutes
* Team 2 joins and dicks around.
[...]

PR #166, opened 2017-07-19
Fixed readycheck system to only use one timer instead of a timer per team.

grenadecx:
Fixes the issue #160.

I removed the g_ReadyTimeWaitingUsed[MatchTeam_Count] and replaced it with g_ReadyTimeWaitingUsed. I didn't see that there was any use for the seperate timer per team instead of one(unless it was planned for future release). 

And hopefully I didn't break anything else in the process.

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -150,7 +150,7 @@ bool g_TeamReadyForUnpause[MatchTeam_Count];
 bool g_TeamGivenStopCommand[MatchTeam_Count];
 int g_TeamPauseTimeUsed[MatchTeam_Count];
 int g_TeamPausesUsed[MatchTeam_Count];
-int g_ReadyTimeWaitingUsed[MatchTeam_Count];
+int g_ReadyTimeWaitingUsed = 0;
 char g_DefaultTeamColors[][] = {
     TEAM1_COLOR, TEAM2_COLOR, "{NORMAL}", "{NORMAL}",
 };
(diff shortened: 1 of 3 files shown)


scripting/get5.sp
```diff
@@ -150,7 +150,7 @@ bool g_TeamReadyForUnpause[MatchTeam_Count];
 bool g_TeamGivenStopCommand[MatchTeam_Count];
 int g_TeamPauseTimeUsed[MatchTeam_Count];
 int g_TeamPausesUsed[MatchTeam_Count];
-int g_ReadyTimeWaitingUsed[MatchTeam_Count];
+int g_ReadyTimeWaitingUsed = 0;
```
Technoblazed: As this is defined globally, the int would default to 0.

```
int g_ReadyTimeWaitingUsed;
```

would suffice.

Technoblazed:
I stand corrected, it's fine how it is.

splewis:
Thanks for the PR!

Looks fine, but I want to test this first. I'm a little concerned about [this section](https://github.com/splewis/get5/blob/master/scripting/get5.sp#L626) being impacted - I think it may need a change if both teams aren't readied to report the match differently somehow.

That said, I don't think that has to be done by you or in this PR (this problem exists right now anyway); I just want to think about what to change here before merging.

grenadecx:
No problem at all. The current problem that exists and did so before I made changes is if both teams are not ready, the first team forfeits the match and makes the second team the winner.

This is something I need to fix as well. In regards to this, the other pull request (#167) from @gabrielfgularte seems to implement some kinda of EndSeriesWithoutWinners. Without taking a closer look at it, maybe that would be feasible to take advantage of if that one is accepted?

miped:
See my comment on issue #160

grenadecx:
This PR is no longer relevant, see issue #160

Issue #165, opened 2017-07-17
Detect player disconnect


gabrielfgularte:
If all players leave the game, it stays `live` and does not finish. On the first player return, the match throw a `round_end` event and the match goes back to the very first round after the knife with 0x0 score. Is this the expected behaviour?

If this is the expected behaviour, is it possible to intercept some disconnect player event? So we can track the players disconnect and/or reconnect actions, and that would be easy to advert, warn and punish the players who deserts.

Besides that, would be nice to have a cVar that finishes the game if the last player leaves, or to configure "what to do if any player disconnect". Things like pause game, end game, restart on reconnect, etc.

splewis:
I haven't put much thought into what happens if everyone leaves a live match, so I'm not sure there's really any expected behavior. Server hibernation & csgo is... annoying.

I'd be fine with some basic behavior like:
- detecting if everyone left, and doing a match restore from the backup when someone came back.
- or ending the match

I guess a cvar for both, and a disabled option (and a forward for this as well).

That said... I don't think I'll get around to this soon.

gabrielfgularte:
@splewis Thank you for your answer.

We need this for our project, so I took it upon myself to make something that could be a starting point. I'll submit a PR for you. Hope it helps.

Even if you don't plan on using it, I would be greatful if you could give me some feedback.

PR #167, opened 2017-07-19
Player disconnect

gabrielfgularte:
Related to #165 
- Add `.DS_Store` MacOS file to ignored files;
- Add an event hook for `PlayerDisconnect` and write it to `EventLogger` (`EventLogger_PlayerDisconnect`);
- Add a var `g_ClientsConnected` that is incremented on OnClientAuthorized and decremented on Event_PlayerDisconnect;
- Add a boolean cVar `get5_end_match_when_everyone_leaves` that finishes the match when all clients disconnect.
- Add an event and a handler for a specific "end series without winners" situation (`EventLogger_SeriesEndWithoutWinners` and `Get5_OnSeriesResultWithoutWinners`)


scripting/get5.sp
```diff
@@ -521,6 +522,12 @@ public Action Event_PlayerConnectFull(Event event, const char[] name, bool dontB
   }
 }
 
+public Action Event_PlayerDisconnect(Event event, const char[] name, bool dontBroadcast) {
+  int client = GetClientOfUserId(event.GetInt("userid"));
+  int site = event.GetInt("site");
```
Technoblazed: Why are you getting the site? On a player disconnect o.O

scripting/get5/eventlogger.sp
```diff
@@ -216,3 +216,11 @@ public void EventLogger_BombExploded(int client, int site) {
   set_json_int(params, "site", site);
   EventLogger_EndEvent("bomb_exploded");
 }
+
+public void EventLogger_PlayerDisconnect(int client, int site) {
+  EventLogger_StartEvent();
+  AddMapData(params);
+  AddPlayer(params, "client", client);
+  set_json_int(params, "site", site);
```
Technoblazed: Same as above, site on disconnect?

scripting/get5.sp
```diff
@@ -74,6 +74,7 @@ ConVar g_TimeFormatCvar;
 ConVar g_VetoConfirmationTimeCvar;
 ConVar g_VetoCountdownCvar;
 ConVar g_WarmupCfgCvar;
+ConVar g_EndMatchOnEveryoneLeavesCvar;
```
Technoblazed: Alphabetically ordered or riot.

gabrielfgularte: Oops, sorry. Fixed.

scripting/get5.sp
gabrielfgularte: My bad. Mistakes of ctrl+c and ctrl+v. Fixed.

scripting/get5/eventlogger.sp
gabrielfgularte: Fixed at b1f0e61 too.

scripting/get5.sp
```diff
@@ -417,6 +426,8 @@ public void OnPluginStart() {
   g_OnSeriesInit = CreateGlobalForward("Get5_OnSeriesInit", ET_Ignore);
   g_OnSeriesResult =
       CreateGlobalForward("Get5_OnSeriesResult", ET_Ignore, Param_Cell, Param_Cell, Param_Cell);
+  g_OnSeriesResultWithoutWinners =
```
splewis: If you add a forward you should update the include (scripting/include/get5.inc). 

That said, I don't think this needs a new forward. I think just calling the existing ``Get5_OnSeriesResult`` with the 1st argument (seriesWinner) as MatchTeam_TeamNone.

gabrielfgularte: But `Get5_OnSeriesResult` expects to receive `team1MapScore` and `team1MapScore` which in this case are both 0 because there's no winner. Also, it gets the stats, which does not exists too.

Do you think that this `Get5_OnSeriesResultWithoutWinners` function is an overkill?

splewis: I still think it should just be one forward. A cancellation is still a result for the series (with 0 map scores).

I'm not sure what you mean by "gets the stats".

gabrielfgularte: I don't know if I NEED to create a forward for the new function (`Get5_OnSeriesResultWithoutWinners`) or if I can write it and use it without a forward. Maybe this is your point: show me that I can use the new function without a forward because it servers to other things. If you're saying that, I understand and I'm going to remove the forward.

But if you're saying that the `Get5_OnSeriesResultWithoutWinners` is too much and `Get5_OnSeriesResult` can handle this situation, let me try to explain in a more detailed way.

The `Get5_OnSeriesResult` function expects a `seriesWinner `, a `team1MapScore` and  a `team2MapScore`. Also, it retrieves the match stats with `Get5_GetMatchStats(kv);`, right? If so, it tests if there was a forfeit to keep this stats or throw them away. After that, the function creates the request to the API.

My point is that if there's no match (pretty much like a forfeit), there's no winners and no stats. So why check if was a forfeit if it wasn't? Same idea for match stats, why try to get the stats if there's no stats for that match?

splewis: > So why check if was a forfeit if it wasn't? Same idea for match stats, why try to get the stats if there's no stats for that match?

I don't understand the problem. If you're trying to optimize the code, the cost of the extra forward and code maintenance massively outweighs any trivial performance change. 

I would still use the existing forward, and pass MatchTeam_TeamNone as the winner.

gabrielfgularte: Hi, sorry for the (giant) delay. If the Forward isn't  necessary, can you help me to remove it? I don't know exactly what I have to remove.

I think that the code might be different by now, considering that this PR is from July. Can you help me to merge the things out and make a valid and updated PR?

Thanks.

gabrielfgularte:
Merged with the updated master branch. :)

gabrielfgularte:
@splewis Any thoughts on this one?

Thanks!

splewis:
Sorry I haven't gotten around to this yet. I'll try to look at it this weekend.

scripting/get5_apistats.sp
```diff
@@ -353,6 +353,17 @@ public void Get5_OnSeriesResult(MatchTeam seriesWinner, int team1MapScore, int t
   g_APIKeyCvar.SetString("");
 }
 
+public void Get5_OnSeriesResultWithoutWinners() {
```
splewis: This forward doesn't exist. Why is this change even needed? ``Get5_OnSeriesResult`` already sends a request, it will just set the http "winner" param to "none".

scripting/get5.sp
```diff
@@ -173,6 +174,9 @@ bool g_PendingSideSwap = false;
 Handle g_KnifeChangedCvars = INVALID_HANDLE;
 Handle g_MatchConfigChangedCvars = INVALID_HANDLE;
 
+/** Informational data **/
+int g_ClientsConnected = 0; // Counting on OnClientAuthorized and Event_PlayerDisconnect are fired only once for each client
```
splewis: I don't see the value in this variable - I'd rather a function just count the number of clients connected instead of maintaining this count.

scripting/get5_apistats.sp
gabrielfgularte: I've removed this. I understand now that `Get5_OnSeriesResult` do the work.I've removed this. I understand now that `Get5_OnSeriesResult` do the work.

scripting/get5.sp
gabrielfgularte: I don't know where to place this function and how to get the number of connected clients. Could anyone help me on this?

boogie77: This could be usefull:
https://github.com/splewis/get5/blob/69f6135548dfbbf199f6356691cee69199729dbf/scripting/get5/teamlogic.sp#L225

scripting/get5.sp
```diff
@@ -534,6 +543,18 @@ public Action Event_PlayerConnectFull(Event event, const char[] name, bool dontB
   }
 }
 
+public Action Event_PlayerDisconnect(Event event, const char[] name, bool dontBroadcast) {
+  int client = GetClientOfUserId(event.GetInt("userid"));
+  g_ClientsConnected--;
+  EventLogger_PlayerDisconnect(client);
+
+  if (g_ClientsConnected <= 0) {
```
splewis: Now that I look at this again, I'd worry about this firing during map changes. Have you tested to make sure that doesn't happen? I guess you could also check to make sure g_MapChangePending if that's an issue.

splewis:
This has been obsoleted by recent commits.

PR #178, opened 2017-08-18
Attempt to resolve last round backup of each map

AlexBksiad:
Before, the last round of each map would write the map number as 2 if the current map was actually still map 1. This is because the way it calculates map# uses the series scores of each team. To fix this we just write the backup prior to the series score being incremented.

The issue this causes if we do not fix it is that, say map 1 goes 23 rounds and map 2 goes 25 rounds, the backup will be written over as the filename would be the same.


splewis:
Sorry I haven't gotten around to this yet. I wanted to do a proper test of this first, I should get around to that in the next days.

PR #198, opened 2017-10-29
Deprecated sv_alltalk

nikolya-prodigy:
"sv_alltalk" = "0" game notify replicated                                        - Deprecated. Replaced with sv_talk_enemy_dead and sv_talk_enemy_living.

### DIFF
--- a/cfg/get5/warmup.cfg
+++ b/cfg/get5/warmup.cfg
@@ -10,7 +10,7 @@ mp_spectators_max 20
 mp_startmoney 16000
 mp_timelimit 0
 mp_warmuptime_all_players_connected 0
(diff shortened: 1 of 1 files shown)


Technoblazed:
Well you might as well add `sv_talk_enemy_dead` as well, no point in just adding the one.

splewis:
This change should be in both the warmup and live configs, and set both the sv_talk_enemy_living and sv_talk_enemy_dead cvars.

splewis:
Closing this since it's missing some cvars - I'm still okay with the change, but it doesn't seem wise to do a partial replacement of the cvars.

PR #200, opened 2017-11-13
Indicate the reason a round ended

damiencornu:
Get the reason a round ended as indicate in SourceMod: [https://sm.alliedmods.net/new-api/cstrike/CSRoundEndReason](https://sm.alliedmods.net/new-api/cstrike/CSRoundEndReason) (1 indexed) so the reason can be retrieved in the `round_end` event and `Get5_OnEvent` hook.
[Events log section](https://github.com/splewis/get5/wiki/Event-logs#map-flow) should be updated if this pull request is merged to show the param is availabe.


splewis:
Looks good to me. And thanks for the reminder on updating the wiki :)

PR #225, opened 2018-02-05
Fix typo in utils.sp

Technoblazed:


splewis:
Based on the uses I see (get5_creatematch, get5_scrim, player name forcing files) this shouldn't have any compatibility issues. 

Thanks!

PR #231, opened 2018-02-22
Add cvar to change hostname to match teams

Technoblazed:
Still needs testing but no reason why this shouldn't work.


splewis:
I don't think this is really worth it as-is. 

If the cvar was formatted like several other cvars get [templated](https://github.com/splewis/get5/blob/master/scripting/get5.sp#L1319) I think this cvar would be much more useful. (also: I think a {SERVERID} param should be in the list of variables you can use in that list)

Technoblazed:
that seem like what you wanted?

splewis:
Yup, thanks!

PR #237, opened 2018-03-27
Admin pause changes

Technoblazed:
Admin pauses should ignore the rest of the function, it will error out later down the function.


splewis:
Thanks!

PR #281, opened 2018-07-19
Display veto information to gotv uses, close #279

### LINKED ISSUE #279: Display map veto information in more places
If you're on GOTV, you may not see what happened during the veto. I don't know what can be changed to make that easier.

Clients in spectate (or admins) could have an additional command to expose the veto information. get5_status currently only shows the maps in the series.

Technoblazed:


scripting/get5.sp
```diff
@@ -999,7 +1004,13 @@ public Action Timer_NextMatchMap(Handle timer) {
...
   char map[PLATFORM_MAX_PATH];
   g_MapsToPlay.GetString(index, map, sizeof(map));
-  ChangeMap(map);
+
+  if (g_DisplayGotvVeto.BoolValue) {
+    float minDelay = float(GetTvDelay()) + MATCH_END_DELAY_AFTER_TV;
+    ChangeMap(map, minDelay);
+  } else {
+    ChangeMap(map);
+  }
```
splewis: Won't this also trigger at the end of map 1 (in a bo3, for example?)

I think the ``minDelay`` variable should be changed in the Event_MatchOver above that creates the timer for ``Timer_NextMatchMap``. Otherwise the delay value is being computed additively in two places, which is needlessly confusing, imo.

scripting/get5.sp
```diff
@@ -963,7 +968,7 @@ public Action Event_MatchOver(Event event, const char[] name, bool dontBroadcast
       g_MapChangePending = true;
       Get5_MessageToAll("%t", "NextSeriesMapInfoMessage", nextMap);
       ChangeState(Get5State_PostGame);
-      CreateTimer(minDelay, Timer_NextMatchMap);
+      CreateTimer(0.1, Timer_NextMatchMap);
```
splewis: Unfortunately my suggestion had an issue. This won't work. Triggering Timer_NextMatchMap too early will cause the demo to stop too early, so the delay probably does have to be added in two places. Sorry for the bad suggestion.

scripting/get5.sp
```diff
@@ -999,7 +1004,13 @@ public Action Timer_NextMatchMap(Handle timer) {
   int index = GetMapNumber();
   char map[PLATFORM_MAX_PATH];
   g_MapsToPlay.GetString(index, map, sizeof(map));
-  ChangeMap(map);
+
+  if (index > 1 || (g_DisplayGotvVeto.BoolValue && index == 1)) {
```
splewis: Shouldn't this be ``== 0``? GetMapNumber is calculated as the sum of map scores, which is certainly 0 right after the veo.

scripting/get5.sp
splewis: This is still a behavior change when g_DisplayGotvVeto.BoolValue is false. The ChangeMap call in this function never passed extra delay before, and now it will pass delay when index > 1 now.

scripting/get5.sp
```diff
@@ -260,6 +261,8 @@ public void OnPluginStart() {
                    "If set to 0, get5 will not force players to the correct team based on steamid");
   g_DemoNameFormatCvar = CreateConVar("get5_demo_name_format", "{MATCHID}_map{MAPNUMBER}_{MAPNAME}",
                                       "Format for demo file names, use \"\" to disable");
+  g_DisplayGotvVeto = CreateConVar("get5_display_gotv_veto", "0",
+                                      "Whether to wait for map vetos to be printed to GOTV before changing map, use \"\" to disable");
```
splewis: Remove the 
>, use \"\" to disable

part. It's misleading since the default is 0 already, and there's lots of other bool cvars that control 'whether' something happens, and how to disable it isn't part of the description.

splewis:
FYI: this looks fine now, but I'm holding off merging until I get a chance to test it.

I also haven't checked really, but it might need some extra checks in the conditional for when a backup is restored.

scripting/get5.sp
```diff
@@ -999,7 +1004,13 @@ public Action Timer_NextMatchMap(Handle timer) {
   int index = GetMapNumber();
   char map[PLATFORM_MAX_PATH];
   g_MapsToPlay.GetString(index, map, sizeof(map));
-  ChangeMap(map);
+
+  if (g_DisplayGotvVeto.BoolValue && index == 0) {
```
splewis: Let's also add a !g_SkipVeto in there as well and I'm happy to merge this.

PR #310, opened 2018-10-14
Update ReadMe

TandelK:
Link reference to Official CVARs of CSGO Dedicated Server by Valve Developers.


splewis:
I think this section should be tweaked a little since it's about both the get5 and built-in csgo cvars.

Maybe a subsection (3 or 4 #'s) for "Get5 ConVars" with the existing content then a new section for builtin cvars that mentions you can customize your live cfg or use the "cvars" section?

TandelK:
I would recommend actually editing the stuff in the Wiki instead of ReadMe. We can have a small brief ConVars which are important for Get5 gameplay specially in a competitive modes.

PR #347, opened 2019-02-10
Fix JSON maplist

Cretezy:
So I'm not too sure how anyone was using the `maplist` feature in JSON before.

- Fix checking for key
- Fix cleaning up (would give `[SM] Exception reported: Invalid Handle` because it doesn't exist in memory)

### DIFF
--- a/scripting/get5/jsonhelpers.sp
+++ b/scripting/get5/jsonhelpers.sp
@@ -75,7 +75,7 @@ stock float json_object_get_float_safe(JSON_Object json, const char[] key, float
 // Used for parsing an Array[String] to a sourcepawn ArrayList of strings
 stock int AddJsonSubsectionArrayToList(JSON_Object json, const char[] key, ArrayList list,
                                        int maxValueLength) {
-  if (json_has_key(json, key, Type_Object)) {
+  if (!json_has_key(json, key, Type_Object)) {
     return 0;
   }
 
@@ -91,7 +91,6 @@ stock int AddJsonSubsectionArrayToList(JSON_Object json, const char[] key, Array
       count++;
     }
     array.Cleanup();
-    delete array;
   }
   return count;
 }
--- a/scripting/get5/matchconfig.sp
+++ b/scripting/get5/matchconfig.sp
@@ -506,7 +506,7 @@ static bool LoadMatchFromJson(JSON_Object json) {
   }
 
   if (AddJsonSubsectionArrayToList(json, "maplist", g_MapPoolList, PLATFORM_MAX_PATH) <= 0) {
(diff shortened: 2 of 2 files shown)


scripting/get5/matchconfig.sp
```diff
@@ -392,7 +392,7 @@ static bool LoadMatchFromKv(KeyValues kv) {
   }
 
   if (AddSubsectionKeysToList(kv, "maplist", g_MapPoolList, PLATFORM_MAX_PATH) <= 0) {
-    LogMessage("Failed to find \"maplist\" section in config, using fallback maplist.");
+    MatchConfigFail("Failed to find \"maplist\" section in config, using fallback maplist.");
     LoadDefaultMapList(g_MapPoolList);
```
splewis: I'm not sure I know what the precise error is yet (I haven't looked through the existing code yet), but I don't think this is the right change. 

The `maplist` list is meant to be optional, `MatchConfigFail` is only meant to be called when it explicitly fails.

Cretezy: Ah I see, would `PrintToServer` be of better use?

splewis: I'm not sure what the problem with LogMessage is here. 

> * Fix output (translation doesn't exist, use standard output)

Is there something in error logs that make you think that? I don't see any issue with the current usage.

scripting/get5/jsonhelpers.sp
```diff
@@ -91,7 +91,6 @@ stock int AddJsonSubsectionArrayToList(JSON_Object json, const char[] key, Array
       count++;
     }
     array.Cleanup();
-    delete array;
```
splewis: Are you sure the invalid handle warning you saw came from this? Unless I'm mistaken, sourcepawn doesn't error on deleting a null handle.

Cretezy: Yep, I was surprised too

scripting/get5/matchconfig.sp
Cretezy: Yeah I think you're right, I'll revert this change

scripting/get5/jsonhelpers.sp
splewis: I'm still not convinced this is a correct change though. Do you have error logs pointing specifically to an invalid handle at this line?

Cretezy: I'll pull the exact error out. But it was definitely an invalid handle.

Cretezy:
```

splewis: Looks like the `GetObject` call `array` comes from returns the internal JSON handle, not a copy of it like I thought. Thanks.

Issue #364, opened 2019-04-10
Invalid Match JSON


PhlexPlexico:
If making a feature request, you should delete all the pre-filled text here.
If reporting a bug, fill in the following sections:

### Expected behavior
Get 5 successfully loads match

### Actual behavior
json_decode: expected key string at 75 when trying to load match from URL.

### Steps to reproduce

Please note that "latest" is NOT a version! See https://github.com/splewis/get5/wiki/Debugging-help#finding-get5-version if you need help finding these. Adding the output file produced by ``get5_debuginfo`` is extremely helpful.

- Plugin version: 0.7.1+458
- Sourcemod version: 1.9
- Steps to reproduce (please be specific):
Use the get5-webpanel to load the match config, receive error in console on server.

Just reporting this from issue https://github.com/splewis/get5-web/issues/147. Seems there's an issue parsing out the JSON with a "+Build#" version?

I know the previous one was reported, but here's a proper bug report and apparently how to reproduce.

splewis:
Thanks.

I confirmed the 0.7.1 release works fine.

I'm guessing the latest dev build has some of the changes I see in the sm-json repository since then: https://github.com/clugg/sm-json/commits/master

I haven't confirmed that's the source of the issue, though.

thboss:
I recompiled plugin for latest json version
get this error
L 04/21/2019 - 00:21:50: [get5.smx] json_decode: couldn't extract string at 631
L 04/21/2019 - 00:21:50: [get5.smx] Failed to load match config: invalid match json
plugin version 0.7.1
SM 1.10

splewis:
I think this has been fixed now? Feel free to reopen if not.

s1inche:
Have you found a solution to this problem?

PhlexPlexico:
You have to make sure you have the SteamWorks extension enabled if you're loading the match remotely. If you don't have all the requirements for the plugin, it will most likely fail.

s1inche:
All plugins work
get5_apistats.smx / G5WS.smx
get5.smx
But when creating a match, it gives this error.

Maybe show you some configs? I just don't understand what to do

PhlexPlexico:
Do you have the SteamWorks extension installed though? `sm exts list`

s1inche:
![image](https://user-images.githubusercontent.com/54583646/165384520-5e5ae15e-7b8d-4af4-8657-b44e634264e0.png)
It seems to be, but I don't see SteamWorks here, is that true?

[...]

PR #367, opened 2019-04-13
Adjusted the SetPlayer function to force name change.

PhlexPlexico:
It is now reloading the playernames in the given kV list, so users names can be custom once they are added, without having to reinit the match, or not have them load at all.

Tested with adding one player to a team after create match. Console reports sv_load_forced_client_names_file: loaded 1 client name(s) to enforce!

### DIFF
--- a/scripting/get5/natives.sp
+++ b/scripting/get5/natives.sp
@@ -141,10 +141,21 @@ public int Native_SetPlayerName(Handle plugin, int numParams) {
   char name[MAX_NAME_LENGTH];
   GetNativeString(1, auth, sizeof(auth));
   GetNativeString(2, name, sizeof(name));
+  char nameFile[] = "get5_names.txt";
+  KeyValues namesKv = new KeyValues("Names");
   char steam64[AUTH_LENGTH];
   ConvertAuthToSteam64(auth, steam64);
   if (strlen(name) > 0 && !StrEqual(name, KEYVALUE_STRING_PLACEHOLDER)) {
     g_PlayerNames.SetString(steam64, name);
+    // Create temp namesKv to only load one name at a time, 
+    // instead of loading in every PlayerName.
+    namesKv.SetString(steam64, name);
+    DeleteFile(nameFile);
+    if (namesKv.ExportToFile(nameFile)) {
+      ServerCommand("sv_load_forced_client_names_file %s", nameFile);
+    } else {
+      LogError("Failed to write names keyvalue file to %s", nameFile);
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


splewis:
I suspect it would be better to change `Get5_SetPlayerName` itself to do this, but I haven't looked through all the callsites quite yet.

Ideally another plugin calling `Get5_SetPlayerName` during a match would still work.

PhlexPlexico:
Sounds good! I've just made the change, but have yet to test it. Through some searching, I don't believe there's another way to force client names without loading from a file... Is this assumption correct? If so, I've basically just ripped a portion of the `LoadPlayerNames()` call and set it in the `Native_SetPlayerName()` call, more or less creating/reading the file from the given name/auth. I don't want to push anything since I haven't tested it yet, but would this be a safe way of approaching it then?

PhlexPlexico:
Tested through rcon commands, it successfully changed the player on `get5_addplayer steamId team1 "newName"` worked. However, the problem from the webpanel still exists (#364).

PhlexPlexico:
Welp, just tested, Yeah definitely a no-go. Now how bad would it be if we just called LoadPlayers(); within that function as well, since we do need to cycle through the player list? Sorry for being so wishy-washy about this, haha. First time attempting to PR to another person's repo.

PhlexPlexico:
Instead of a lazy fix, I opted to check if the file exists on the server, reload it to a KeyValue pair, and swap the steam ID with whichever name is given. This way, if this file remains persistent, it will reload to whatever the latest name should be. I tested with a few different players loading in, removing them from the game, and adding them with new names, as well as loading one individual name, and then deleting the file afterwards as well.

scripting/get5/natives.sp
```diff
@@ -139,15 +139,34 @@ public int Native_AddPlayerToTeam(Handle plugin, int numParams) {
 public int Native_SetPlayerName(Handle plugin, int numParams) {
   char auth[AUTH_LENGTH];
   char name[MAX_NAME_LENGTH];
+  KeyValues namesKv = new KeyValues("Names");
+  char nameFile[] = "get5_names.txt";
```
splewis: I think this approach has some problems.

1. In terms of code health, there is duplicate code with this function and `LoadPlayerNames` now. I don't see a clear reason why there needs to be two places that implement this code.

2. If multiple server processes are running on the server installation the `get5_names.txt` files will clash and you'll get a weird file.

Barring some unknown valve server bug, I don't see any clear reason to try to be efficient with the file building/loading.

PhlexPlexico: Sounds good! Just wasn't sure which approach would be best (reloading everyone's name vs. appending and updating). I'll revert the last few commits to just call `LoadPlayerNames()` inside the `SetPlayerName` call then :)

PR #370, opened 2019-04-15
Absolute ready timer

rgoupil:
Sounds related to #160.

This PR replace the two independent time pools by an absolute one in order to bring a more consistent behaviour as well as preventing any stalling with the ready time.

Here are the problems we encountered with the original logic:
* If both teams ready up and then unready, the team that was ready for the most time is declared as winner. This is odd for the players as the winner seem to be randomly picked, it sound logical that both team should be forfeited
* With ready check set to N seconds, it is technically possible to stall the match for (N * 2) - 1 seconds. If you also take into account that the ready check seem to be reset when the first player join the server, you can go up to (N * 3) - 1
* If both team don't ready up, Team 1 is declared as winner. This gives an advantage to a team that was not ready as well as not properly detecting the case where both team do not show up - this was a killer for us

The single absolute timer fixes all of these problems by simply checking that after the specified time any team that is not ready is forfeited, no matter how long they were ready for or if they have been ready before.

While I can understand that replacing a behaviour is potentially annoying for existing setups, I would argue that the previous logic had several flaws while arguably not bringing much (except the feature itself of course).

Thanks for Get5, we really like it over here and though we could help to make it better ;)

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -697,8 +697,7 @@ static void CheckReadyWaitingTimes() {
     if (team1Forfeited && team2Forfeited)
     {
       g_ForcedWinner = MatchTeam_TeamNone;
-      Stats_Forfeit(MatchTeam_Team1);
-      Stats_Forfeit(MatchTeam_Team2);
+      Stats_Forfeit(MatchTeam_None);
     }
     else if (team1Forfeited)
     {
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -693,8 +693,7 @@ static void CheckReadyWaitingTimes() {
 
     if (team1Forfeited && team2Forfeited) {
       g_ForcedWinner = MatchTeam_TeamNone;
-      Stats_Forfeit(MatchTeam_Team1);
-      Stats_Forfeit(MatchTeam_Team2);
+      Stats_Forfeit(MatchTeam_TeamNone);
     } else if (team1Forfeited) {
       g_ForcedWinner = MatchTeam_Team2;
       Stats_Forfeit(MatchTeam_Team1);
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -158,7 +158,7 @@ bool g_TeamGivenStopCommand[MatchTeam_Count];
 bool g_InExtendedPause;
 int g_TeamPauseTimeUsed[MatchTeam_Count];
 int g_TeamPausesUsed[MatchTeam_Count];
-int g_ReadyTimeWaitingUsed[MatchTeam_Count];
+int g_ReadyTimeWaitingUsed = 0;
 char g_DefaultTeamColors[][] = {
     TEAM1_COLOR, TEAM2_COLOR, "{NORMAL}", "{NORMAL}",
 };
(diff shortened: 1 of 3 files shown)


splewis:
I agree with your analysis - and that the current approach has some problems.

I'll try to take a better look at this soon, but I would expect this to get merged. Thanks for the contribution!

scripting/get5.sp
```diff
@@ -686,35 +686,57 @@ public Action Timer_CheckReady(Handle timer) {
 
 static void CheckReadyWaitingTimes() {
   if (g_TeamTimeToStartCvar.IntValue > 0) {
-    CheckReadyWaitingTime(MatchTeam_Team1);
-    CheckReadyWaitingTime(MatchTeam_Team2);
+    bool team1Forfeited;
+    bool team2Forfeited;
```
splewis: These should just be declared where they're set a few lines down.

scripting/get5.sp
```diff
@@ -686,35 +686,57 @@ public Action Timer_CheckReady(Handle timer) {
...
+
+    g_ReadyTimeWaitingUsed++;
+
+    team1Forfeited = CheckReadyWaitingTime(MatchTeam_Team1);
+    team2Forfeited = CheckReadyWaitingTime(MatchTeam_Team2);
+
+    if (team1Forfeited && team2Forfeited)
+    {
```
splewis: style nitpick: to be consistent with the rest of the code, can you format this in the 
```
if (...) {
  ...
} else if (...) {
  ...
}
```

style?

rgoupil:
We are resuming work on CSGO next week, I'll get on the styling issues monday, thanks :)

rgoupil:
Done, please let me know if anything else looks off to you.

grenadecx:
I never got around to make a PR for this, but I'm glad someone did. I still think this makes the most sense.

scripting/get5.sp
```diff
@@ -686,35 +686,49 @@ public Action Timer_CheckReady(Handle timer) {
...
-    CheckReadyWaitingTime(MatchTeam_Team1);
-    CheckReadyWaitingTime(MatchTeam_Team2);
+    g_ReadyTimeWaitingUsed++;
+
+    bool team1Forfeited = CheckReadyWaitingTime(MatchTeam_Team1);
+    bool team2Forfeited = CheckReadyWaitingTime(MatchTeam_Team2);
+
+    if (team1Forfeited && team2Forfeited) {
+      g_ForcedWinner = MatchTeam_TeamNone;
+      Stats_Forfeit(MatchTeam_Team1);
```
splewis: I think multiple calls to Stats_Forfeit yields misleading resulting.

As written, this will end up calling `Stats_SeriesEnd` for both teams and `STAT_SERIESWINNER` will be one of the teams. It should probably be mapping to the none team in that case.

rgoupil:
Is 555f3c46bb906b9e681e89d90bf34352b2f68f85 what you wanted @splewis ?

rgoupil:
I just realised this might lead to an issue where it is not possible to tell the difference between a match ended on a double forfeit or a match ended by an admin, since in both cases, the `OnSeriesResult`, `Stats_SeriesEnd`, etc. are called with `MatchTeam_TeamNone` as the winning team.

Let me know if I missed something, we'll be looking for a solution on our end as we would like our plugin to be able to tell between the two.

splewis:
Looks good to me, thanks.

PR #383, opened 2019-05-09
Pause and Unpause events and autoready after map veto

gamefraym:
Created since we wanted to handle an event created from pause commands from our extension and enforce auto ready for all players after map veto.


splewis:
Can you split this into 2 pr's? I'm not sure how I feel about the auto ready, but the pausing is definitely worth merging.

gamefraym:
Will split it in two and discuss the autoready in its respective.

PR #386, opened 2019-05-10
Pause and Unpause events

gamefraym:
Pause events.


splewis:
Hm, I'm not sure this works as well as I'd like.

With this, if a team does a 30-second fixed-time pause the "unpause" event will never happen. 

This change *might*   be okay if the events were "pause_command" and "unpause_command" - but I think at least passing the parameters of the pause would be useful for the "pause_command" event in that case.

I think it'd probably still be preferable if the "unpause"  event actually fired when the game actually unpaused, though.

Drarig29:
Hi @splewis, this can now be closed I think 😊

PhlexPlexico:
This PR can be closed now since it has been implemented.

PR #387, opened 2019-05-10
Added auto ready after map veto config option.

gamefraym:
This option is added due to the following flow when having map veto: players join and have 10 minutes to ready up. Veto commences and map changes. Then there is 10 new minutes where the player has to ready up again. We want a fast-track approach where we expect every player to be ready after veto.

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -81,6 +81,7 @@ ConVar g_TimeFormatCvar;
 ConVar g_VetoConfirmationTimeCvar;
 ConVar g_VetoCountdownCvar;
 ConVar g_WarmupCfgCvar;
+ConVar g_AutoReadyAfterVetoCvar;
 
 // Autoset convars (not meant for users to set)
 ConVar g_GameStateCvar;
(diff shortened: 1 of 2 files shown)


gamefraym:
Is there any reason why CreateVeto() does ResetReadyStatus()? 
Can this be configured to be skipped?
Or is it better to store current ready state and set it after veto is complete?

splewis:
How does this relate to #388? Is it two approaches you've found for your problem, or are both desired?

scripting/get5.sp
```diff
@@ -329,7 +330,9 @@ public void OnPluginStart() {
       CreateConVar("get5_veto_countdown", "5",
                    "Seconds to countdown before veto process commences. Set to \"0\" to disable.");
   g_WarmupCfgCvar =
-      CreateConVar("get5_warmup_cfg", "get5/warmup.cfg", "Config file to exec in warmup periods");
+      CreateConVar("get5_warmup_cfg", "get5/warmup.cfg", "Config file to exec in warmup periods");
```
nickdnk, line 333: @gamefraym Please remove these changes as it's formatting only.

PR #388, opened 2019-05-10
Added option to skip resetting ready state on map veto.

gamefraym:
An option to auto ready after map veto is to keep the ready state. This is to prevent users failing to understand that they have to do !ready two times before a match starts causing unwanted forfeits.

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -81,6 +81,7 @@ ConVar g_TimeFormatCvar;
 ConVar g_VetoConfirmationTimeCvar;
 ConVar g_VetoCountdownCvar;
 ConVar g_WarmupCfgCvar;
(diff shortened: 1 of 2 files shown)


splewis:
> This is to prevent users failing to understand that they have to do !ready two times before a match starts causing unwanted forfeits.

I'm not sure I understand what the problem is. Typically the sequence of events is:
1) players join server
2) players !ready
3) map veto happens
4) map change
5) players load
6) players !ready again

Are you saying having to ready in 6) *after a map change* is problematic for you? Or am I missing something?

nickdnk:
@gamefraym Hello and thanks for your contribution.

If you're still interested in having these changes merged, please respond to the comments made on your PRs. If you don't, I will close them.

PR #426, opened 2019-09-08
add ready events

gelzis:
Hello

I added events for team ready/unread. We are not using strict timers and forfeits, but still in times of someone abusing the ready system we are logging it and issuing manual penalties and this helps to get get exact timing of ready/unready.


gelzis:
@splewis can I get any comment on this?

PR #429, opened 2019-09-19
upload team logos to new location for panorama

pikaxo:
as title said its just a correction because panorama

### DIFF
--- a/scripting/get5_apistats.sp
+++ b/scripting/get5_apistats.sp
@@ -43,7 +43,13 @@ ConVar g_APIURLCvar;
 char g_APIURL[128];
 
 //This logo dir its for old csgo hud
-#define LOGO_DIR "materials/panorama/images/tournaments/teams"
+if (g_UseSVGCvar.BoolValue) {
+      #define LOGO_DIR "materials/panorama/images/tournaments/teams"
+    }
+    else{
+      #define LOGO_DIR "resource/flash/econ/tournaments/teams"
+    }
+
 
 // clang-format off
 public Plugin myinfo = {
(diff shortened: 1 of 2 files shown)


splewis:
Can you make this change backwards compatible? I'm willing to change the default to svg and the new paths, but not drop support for existing behavior entirely. A new cvar would be fine.

scripting/get5_apistats.sp
```diff
@@ -35,14 +35,20 @@
...
 char g_APIKey[128];
 
 ConVar g_APIURLCvar;
 char g_APIURL[128];
 
-#define LOGO_DIR "resource/flash/econ/tournaments/teams"
+//This logo dir its for old csgo hud
+    if (g_UseSVGCvar.BoolValue) {
+      #define LOGO_DIR "materials/panorama/images/tournaments/teams"
+    } else {
```
splewis: This doesn't work  (or compile), can you do something like:

```
#define LEGACY_LOGO_DIR  "resource/flash/econ/tournaments/teams"
#define LOGO_DIR  "materials/panorama/images/tournaments/teams"
```

Then you can do

```
  if (g_UseSVGCvar.BoolValue) {
    Format(logoPath, sizeof(logoPath), "%s/%s.svg", LOGO_DIR, logo);
  } else {
    Format(logoPath, sizeof(logoPath), "%s/%s.png", LEGACY_LOGO_DIR , logo);
  }
```

pikaxo: i saw it on same day but it's too late for me

splewis:
Was added in #498

PR #432, opened 2019-09-24
Update get5.phrases.txt - FR Translation

tech62:


splewis:
Please remove the "#format" lines - they should only be in the main English file.

tech62:
I've update the file, told me if it's good ;)

Issue #393, opened 2019-05-22
Depend on sm-json as a git submodule


splewis:
Versioning would be less problematic if it was included in-tree.

See:
https://github.com/splewis/get5/blob/master/.travis.yml#L27
https://github.com/splewis/get5/blob/master/scripting/get5.sp#L25

splewis:
Done, and fixed jenkins builds. Thanks again @ZeroParticle.

PR #446, opened 2019-12-12
Add sm-json submodule

ZeroParticle:
Implements #393 

Uses the v1.x branch that reverts the breaking change.


splewis:
Thanks! I'll need to update the jenkins build script as well (it's not in the repo, I'll update it after I merge this, when I have time). 

Travis failure for SM 1.10 appears unrelated - it fails at HEAD already. I'll go drop 1.10 testing for now...

PR #460, opened 2020-01-30
Reset g_HasKnifeRoundStarted when knife round ends

AntiPaste:
I'm struggling with a hard-to-replicate bug where the knife round ends before it even starts. I haven't been able to get more debug information than seeing it happen on video yet.

You can see it for yourself here: https://www.twitch.tv/videos/543586511?t=03h56m48s

Last player readies up, the knife winner is announced before the "The knife round will begin in 5 seconds" message and it's just a mess after that.

This seems to happen only when a match has been played, ended, and another match loaded & started without restarting the server. My hunch looking at the code is that there is a race condition due to `g_HasKnifeRoundStarted` not being reset anywhere except in the knife start function.

It might be that having `mp_ignore_round_win_conditions 1` in `warmup.cfg` and `mp_ignore_round_win_conditions 0` in `knife.cfg` triggers the otherwise hidden bug, in some cases.

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -1142,6 +1142,8 @@ public Action Event_RoundEnd(Event event, const char[] name, bool dontBroadcast)
   }
 
   if (g_GameState == Get5State_KnifeRound && g_HasKnifeRoundStarted) {
+    g_HasKnifeRoundStarted = false;
+
     ChangeState(Get5State_WaitingForKnifeRoundDecision);
     CreateTimer(1.0, Timer_PostKnife);
 


AntiPaste:
After running the server for a few days with this PR applied, I can confirm that this fix removes the bug and knife rounds work as expected.

MikkelDK:
Thanks for that @AntiPaste - I will add it to my server ^^

splewis:
Sorry for the delay: I had looked at this a while ago and hoped I could figure out a better way to adjust to code so this state appears more coherent (like could the g_GameState variable/enum be adjusted to make this bool not required), but I'm not sure what to do still, so I'll go ahead and merge since I think this PR is strictly better, solely by probably fixing the bug.

PR #463, opened 2020-02-16
Update README.md

Drarig29:
This line says that `num_maps` must be an "odd number **or** 2":

https://github.com/splewis/get5/blob/2da6d78e46bd4f82b25ee2042a7411480ec76ec6/scripting/get5/matchconfig.sp#L353

And this line says the first team to be CT when `never_knife` is set is team1.

https://github.com/splewis/get5/blob/2da6d78e46bd4f82b25ee2042a7411480ec76ec6/scripting/get5/matchconfig.sp#L90


splewis:
Thanks for the corrections!

PR #466, opened 2020-02-19
Delay replenishing money on spawn

AntiPaste:
Currently giving 16000 to each player on spawn fails so often that I wasn't even aware this is a feature. After Googling the issue I came across a thread that said setting player money on spawn might fail because the player hasn't actually spawned yet when the event is fired and suggested adding a 0.1 second timer on it.

After adding the timer I can confirm that this fixes the issue and money is correctly replenished in warmup upon spawning.


splewis:
Agreed with the fix - nice find.

Issue #450, opened 2020-01-06
get5_kick_immunity "1" not working (everyone can join)


ghost:
### Expected behavior
Allow only admins to join on server with no match loaded.

### Actual behavior
Everyone can join the server with get5_kick_immunity set to "1"

### Steps to reproduce
- Plugin version: "get5_version" = "0.7.2-dev+474"
- Sourcemod version: 1.10.0.6458
- Steps to reproduce (please be specific):

1. Set get5_kick_immunity "1" and get5_kick_when_no_match_loaded "1"
2. Don't add you as admin in admins_simple.ini 
3. Try to connect to passwordless server without a loaded matchconfig
4. You can join server

Server was rebooted on every config change and SM admin was confirmed with executing multiple SM admin commands. The get5.cfg is on all default values.

I also tried setting get5_kick_immunity "0" and then I don't get access with any user, even the user added as admin.

Did https://github.com/splewis/get5/issues/372 break this when trying to fix it?

TandelK:
Even i had the same issue, we did try to modify the code as per new Boolean but it did not work properly as it should. Would also again look into code for it

TandelK:
@splewis i did some digging and found out that 

CheckCommandAccess() is not all working

https://github.com/splewis/get5/blob/0609e51c40e5eeda30dc0ab4f599ab89054a3916/scripting/get5.sp#L534

Please have a look into it.

TandelK:
@splewis Can you check this out.

TandelK:
@Enghausen-petabit-gg here is a fix for it.

TandelK:
@Enghausen-petabit-gg here its fixed

Enghausen-petabit-gg:
Great work! I will test it later this week! Thanks alot!!

Issue #163, opened 2017-07-06
Team names include [NOT READY] tag after backup restoration


jongcs:
### Expected behavior
[NOT READY] in team name after load the last match backup file (!stop).
If team name in match cfg file is "TEAM NAME".
### Actual behavior
After use !stop command, the team name will be:
[NOT READY] TEAM NAME

### Steps to reproduce
- Plugin version: 0.5.2
- Sourcemod version: 1.8.0.6014

splewis:
Can you give clear steps to reproduce?

If this does happen, it should go away once teams ready up after the stop command. Can you clarify if it does?

jongcs:
Game start. -> Pause the game, because someone disconnected. -> Both team use !stop to reload. -> Both team name change to "[NOT READY] TEAM NAME", but they don't need to !ready. -> Both team !unpause the game. -> Game resume.

This has happened on my match, so I changed mp_teamname manual.

Technoblazed:
Actually, something similar to this happens, the teamname doesn't actually remove the  notready part until after the first round ends, at least from my experience, so your !stop would cause it to load the knife config, which has NOT READY in use.

jongcs:
It wasn't first round.

Before !stop:
![image](https://user-images.githubusercontent.com/21971007/28003307-4bf9eac6-656f-11e7-9971-7804e3e04a0e.png)

After !stop:
![image](https://user-images.githubusercontent.com/21971007/28003317-5b4e7118-656f-11e7-8bac-c454f6455bfd.png)

johankh92:
I confirm. This happens after the round is restored.

PhlexPlexico:
So yeah, I think https://github.com/splewis/get5/pull/652 may have fixed this issue, but I'm not entirely sure. Throughout my testing for various things with backups I didn't run into this anymore. Is anyone able to confirm if this is still happening?

PhlexPlexico:
Had some more testing done and it looks like that PR may have fixed it. People are no longer receiving the team names when doing round restores.

PR #477, opened 2020-03-18
CVAR for Team Ready Name Status and Fix Admin Immunity Kick Check

TandelK:
Added CVAR for Team Ready and Not Ready Status in warmup. The reason why i added this is a small workaround for [NOT READY] getting displayed after Round restores. This will disable the part if people dont want to use [READY] [NOT READY] on Team Names in Warmups.

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -531,13 +531,6 @@ public void OnClientAuthorized(int client, const char[] auth) {
     return;
   }
 
-  if (g_GameState == Get5State_None && g_KickClientsWithNoMatchCvar.BoolValue) {
-    if (!g_KickClientImmunity.BoolValue ||
-        !CheckCommandAccess(client, "get5_kickcheck", ADMFLAG_CHANGEMAP)) {
-      KickClient(client, "%t", "NoMatchSetupInfoMessage");
-    }
-  }
-
   if (g_GameState != Get5State_None && g_CheckAuthsCvar.BoolValue) {
     MatchTeam team = GetClientMatchTeam(client);
     if (team == MatchTeam_TeamNone) {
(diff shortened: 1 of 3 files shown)


TandelK:
https://github.com/splewis/get5/issues/163

scripting/get5/util.sp
```diff
@@ -260,15 +260,21 @@ stock void SetTeamInfo(int csTeam, const char[] name, const char[] flag = "",
...
+        !g_DoingBackupRestoreNow) {
+      MatchTeam matchTeam = CSTeamToMatchTeam(csTeam);
+      if (IsTeamReady(matchTeam)) {
+        Format(taggedName, sizeof(taggedName), "%T %s", "ReadyTag", LANG_SERVER, name);
+      } else {
+        Format(taggedName, sizeof(taggedName), "%T %s", "NotReadyTag", LANG_SERVER, name);
+      }
     } else {
-      Format(taggedName, sizeof(taggedName), "%T %s", "NotReadyTag", LANG_SERVER, name);
+      strcopy(taggedName, sizeof(taggedName), name);
```
splewis: This if/else structure seems more confusing than it needs to be. Can you find a way to collapse the 
 strcopy(taggedName, sizeof(taggedName), name);` case (when the tags aren't added) so it's specified once? This code seems like it should only need 3 branches, not 4.

scripting/get5/util.sp
```diff
@@ -260,15 +260,21 @@ stock void SetTeamInfo(int csTeam, const char[] name, const char[] flag = "",
 
   // Add Ready/Not ready tags to team name if in warmup.
   char taggedName[MAX_CVAR_LENGTH];
-  if ((g_GameState == Get5State_Warmup || g_GameState == Get5State_PreVeto) &&
-      !g_DoingBackupRestoreNow) {
-    MatchTeam matchTeam = CSTeamToMatchTeam(csTeam);
-    if (IsTeamReady(matchTeam)) {
-      Format(taggedName, sizeof(taggedName), "%T %s", "ReadyTag", LANG_SERVER, name);
+  if (g_ReadyTeamTag.BoolValue)
+  {
```
splewis: Please follow the bracing style already used in this code (put the brace on the same line as the if)

scripting/get5/util.sp
TandelK: Basically when it is set to 1, The [READY] and [NOT READY] tags need to go out when Warmup is over so in order to bring them to normal structure it was added.

TandelK: If we dont add the last else statement , when the Cvar is set to 0 the server wont even print team names and just print it as Counter Terrorist and Terrorist as the Team Names.

TandelK:
@splewis Admin were getting kicked with No Match Setup Issue as described in issue of https://github.com/splewis/get5/issues/450. This should fix it. OnClientAuthorized did not even had confirmation with Admin checks. Have test all the conditions including overrides and working as it should,

TandelK:
The reason why i have used IsPlayer is because GOTV will be getting kicked as well .

scripting/get5.sp
```diff
@@ -549,6 +544,17 @@ public void OnClientAuthorized(int client, const char[] auth) {
   }
 }
 
+public void AdminClientKickCheck(int client) {
+  if (IsPlayer(client)) {
```
TandelK, line 548: Added because only Players have to be checked otherwise it will also kick Bots which we require specially GOTV for recording

PR #484, opened 2020-03-29
Fixed black screen error when no match setup

b3none:
I'm not really sure why this was changed before but it was just allowing anyone to join the server even when there was no match setup


splewis:
Hmm, yes, this indeed is wrong. Thanks. Came from https://github.com/splewis/get5/pull/427 originally.

PR #487, opened 2020-04-05
Fix Admin Immunity Check

TandelK:
On ClientAuthorized does not check for Admins and they are only configured after OnClientPostAdminCheck(); 

Fix for #450

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -522,17 +522,6 @@ public Action Timer_InfoMessages(Handle timer) {
   return Plugin_Continue;
 }
 
-public void AdminClientKickCheck(int client) {
-  if (IsPlayer(client)) {
-    if (g_GameState == Get5State_None && g_KickClientsWithNoMatchCvar.BoolValue) {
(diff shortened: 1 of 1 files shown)


TandelK:
i have used IsPlayer() is because it will also kicks BOTs from the server leading to GOTV Kick which we dont want.

scripting/get5.sp
```diff
@@ -522,20 +522,24 @@ public Action Timer_InfoMessages(Handle timer) {
...
+public void AdminClientKickCheck(int client) {
+  if (IsPlayer(client)) {
+    if (g_GameState == Get5State_None && g_KickClientsWithNoMatchCvar.BoolValue) {
+      if (!g_KickClientImmunity.BoolValue ||
+          !CheckCommandAccess(client, "get5_kickcheck", ADMFLAG_CHANGEMAP)) {
+            KickClient(client, "%t", "NoMatchSetupInfoMessage");
+      }
+    }
+  }
+}
```
splewis, line 534: Can you just move this into `OnClientPostAdminCheck`? The extra function seems unnecessary.

Issue #495, opened 2020-04-10
Add player score to recorded stats


ringvold:
Thanks for this project! It makes it much easier to run tournaments, even just the casual ones we do every week! 😄 

I cant find anything in the stats file that looks like the player score you can see at right end of the in-game score table  when holding tab. 
The stats file have a lot of other stats but we would like to record the player score to see the performance over time (also the official ranks don't always tell the whole truth). Right now we are taking screenshots and adding it manually to a spreadsheet but since stats are already being written for the matches it would be nice have that score available.

Are there any reason this score is not included or cant it be produces from the stats that are currently being written in the stats files?

splewis:
No reason it's not there (other than me just not caring about the stat at the time), it'd be quite easy to add a new stat. PRs welcome.

ringvold:
I have never done this kind of programming but I'll give it a try. :) Do you have any pointers on where to look? Do you know if it is available to just pull out from the game or if it needs to be calculated?

ringvold:
After some poking around I think this might be the one I'm looking for: https://sm.alliedmods.net/new-api/cstrike/CS_GetClientContributionScore

splewis:
Yup, that should be it. Many of the stats get calculated from hooks (e.g. `player_death` has a callback fired each death, which updates the stats), but that for score it's probably better to just do it on round-end, which you could add to this loop doing some other stats: https://github.com/splewis/get5/blob/master/scripting/get5/stats.sp#L91

PR #497, opened 2020-04-13
Add player contribution score to recorded stats

ringvold:
Adds the player contribution score to recorded stats as discussed in #495.

Also adds the contribution score to apistats post request back to web server.

I chose to use the name `contribution_score` as this seems to be the correct term but most people might be familiar with is as just the player score. I don't have any preferences.

### DIFF
--- a/misc/import_stats.sql
+++ b/misc/import_stats.sql
@@ -93,6 +93,7 @@ CREATE TABLE `get5_stats_players` (
   `firstkill_ct` smallint(5) unsigned NOT NULL,
   `firstdeath_t` smallint(5) unsigned NOT NULL,
   `firstdeath_ct` smallint(5) unsigned NOT NULL,
+  `contribution_score` smallint(5) unsigned NOT NULL,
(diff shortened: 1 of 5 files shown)


saintplay:
I think you also need to update this file:  `/misc/import_stats.sql`

ringvold:
I did not do anything in regards to the mysqlstats plugin, but I guess it makes sense to add it there as well.

jeaye:
Thanks for your work on this, @ringvold! I'm excited to see this happening.

splewis:
Thanks.

Note to myself:

Next release should include a compatibility note for the schema change here. i.e. tell users to run:
```
MODIFY TABLE `get5_stats_players` ADD COLUMN `contribution_score` smallint(5) unsigned;
```

Or something like that. If someone can test to ensure the exact syntax is correct, that would help :)

rdsedmundo:
You forgot a comma here:
https://github.com/splewis/get5/pull/497/files#diff-f14c9391d01dd62f530a5dc944924e78R284

PR #498, opened 2020-04-14
added panorama logo dir

MoritzLoewenstein:
exact replacement for pull request #429 (i dont have write access to #429 ).

### DIFF
--- a/scripting/get5_apistats.sp
+++ b/scripting/get5_apistats.sp
@@ -27,7 +27,7 @@
 #include "get5/version.sp"
 
 #include <SteamWorks>
-#include <json>  // github.com/clugg/sm-json
+#include <json> // github.com/clugg/sm-json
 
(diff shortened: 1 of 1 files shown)


splewis:
Thanks for finishing this up!

tuudik:
Do I understand correctly that the get5 itself still doesnt support svg? @splewis @MoritzLoewenstein 
Also it I've been trying out it and noticed that following code doesnt work:
```
GetConVarStringSafe("mp_teamlogo_1", logo1, sizeof(logo1));
GetConVarStringSafe("mp_teamlogo_2", logo2, sizeof(logo2));
```
I doesnt get the values from json which means it wont try to download them too.

MoritzLoewenstein:
It supports svg (by default enabled), but they need to be in the correct directory (`LOGO_DIR` in the code). If this doesnt work you should open a new issue.

PR #499, opened 2020-04-14
specified sm-json branch

MoritzLoewenstein:
sm-json v3 causes these compilation errors:
```
get5/jsonhelpers.sp(9) : error 139: could not find type "JSON_CELL_TYPE"
get5/jsonhelpers.sp(28) : error 139: could not find type "JSON_CELL_TYPE"
```
another solution would be to adjust jsonhelpers.sp to work with sm-json v3.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "dependencies/sm-json"]
 	path = dependencies/sm-json
 	url = https://github.com/clugg/sm-json
+	branch = v2.x
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,3 @@
 [submodule "dependencies/sm-json"]
 	path = dependencies/sm-json
 	url = https://github.com/clugg/sm-json
-	branch = v2.x
--- a/scripting/get5/jsonhelpers.sp
+++ b/scripting/get5/jsonhelpers.sp
@@ -1,3 +1,5 @@
+#include <string>
+
 stock JSON_Object json_load_file(const char[] path) {
   File f = OpenFile(path, "r");
   char contents[8192];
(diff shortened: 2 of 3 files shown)


TandelK:
Splewis has already used specific commit branch linking to Dependencies which is working perfectly unless people try to just download the latest version for downloading and compiling it. Even i did the same thing and tried it and saw that specific commit linked is working perfectly.

MoritzLoewenstein:
too bad we dont really know the breaking changes in sm-json. i guess this can be closed if we dont want to take the risk of errors, it does compile at least with v2.x.  
I wasnt really familiar with git submodules before, sorry.

splewis:
This PR (as-is) is effectively moving the submodule from being at a specific commit of sm-json to the latest at a specific branch, yes?

I'm not quite sure about that yet, but I would in favor of updating things to work with sm-json at HEAD, in general. Whether that's pinning to a newer commit, or the v3 branch, I'm not quite sure yet. Any thoughts?

TandelK:
@splewis Recently sm-json was updated and seems to have crash Get5 for some reason. If we use the commit linked with dependencies with get5 it works perfectly.

MoritzLoewenstein:
Making get5 compatible with the current sm-json version could potentially be a simple rename ([breaking changes](https://github.com/clugg/sm-json/releases/tag/v3.0.0)) from `JSON_CELL_TYPE` to `JSONCellType` and prefixing the types in `json_string_type` function with `JSON_`. Maybe I will test this later.  
Currently this pr is setting the commit to the newest commit in v2.x branch (and specifying the branch).

MoritzLoewenstein:
I had to remove `Type_Null` from `json_string_type()`,maybe this can cause issues. Everything else is just a replacement with the newest sm-json, no change in functionality.

PR #501, opened 2020-04-18
Fix warning 241 when compiling in SM 1.10

Drarig29:
When I compiled with SourceMod 1.10, I used to get this warning for every line where there is an `ArrayList` or array of size `MatchTeam_Count`.

> Warning 241: Array-based enum structs will be removed in 1.11. See https://wiki.alliedmods.net/SourcePawn_Transitional_Syntax#Enum_Structs

It annoyed me, so I introduced the constant `MATCHTEAM_COUNT` set to `4` and used it everywhere.

I couldn't remove `MatchTeam_Count` from the enum because of the loop in the macro `LoopTeams()`. It's enumerating over the enum, so an element of the enum is also needed in the loop's condition.

Maybe someone will find a better fix and, by the way, maybe using SM 1.10 is not advised? (I didn't find anything about that.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ addons:
         - lib32stdc++6  # needed for spcomp
 
 env:
-    - SMVERSION=1.9
+    - SMVERSION=1.10
 
 before_script:
     # setup submodules
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -96,14 +96,14 @@ int g_MapsToWin = 1;  // Maps needed to win the series.
 bool g_BO2Match = false;
 char g_MatchID[MATCH_ID_LENGTH];
 ArrayList g_MapPoolList = null;
-ArrayList g_TeamAuths[MatchTeam_Count];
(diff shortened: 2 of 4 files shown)


splewis:
Can you add SM 1.10 here as well: https://github.com/splewis/get5/blob/master/.travis.yml#L8. That will automatically test that it compiles without warnings.

Drarig29:
Done.

By the way, it's the same warning (converted to error) as here for example:
https://travis-ci.org/github/splewis/get5/jobs/623932184#L451

MikkelDK:
Seems @splewis need to update the plugin if it needs to being supported for SM 1.11 =)

`./scripting/get5.sp(99) : error 241: Array-based enum structs will be removed in 1.11. See https://wiki.alliedmods.net/SourcePawn_Transitional_Syntax#Enum_Structs`

Drarig29:
> Seems @splewis need to update the plugin if it needs to being supported for SM 1.11 =)

@MikkelDK Actually, this is what this PR does... 😉

PR #506, opened 2020-04-25
Fixing AddPlayerStatus sql and import_stats.sql

ringvold:
As @rdsedmundo mentioned in #497 it was missing a comma in the `AddPlayerStats` in `get5_mysqlstats`. Added tradekill as suggested as well.

Also some formatting to make Format() parameters easier to read.

### DIFF
--- a/misc/import_stats.sql
+++ b/misc/import_stats.sql
@@ -93,6 +93,7 @@ CREATE TABLE `get5_stats_players` (
   `firstkill_ct` smallint(5) unsigned NOT NULL,
   `firstdeath_t` smallint(5) unsigned NOT NULL,
   `firstdeath_ct` smallint(5) unsigned NOT NULL,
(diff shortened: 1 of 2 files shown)


ringvold:
Hmm. Its seems to fail. Not quite understanding why, but I'm looking into it.

rdsedmundo:
Thanks for the fixing it and also adding the tradekill field.

It is failing because you forgot a backslash (`\`) at [get5_mysqlstats.sp#L294](https://github.com/splewis/get5/pull/506/files#diff-f14c9391d01dd62f530a5dc944924e78R294).

ringvold:
Thanks for the help, @rdsedmundo!
Should be fixed now. :smile:

ringvold:
Fixed a commit message. Does not look to great here, but looks fine in the "Commits" tab. :sweat_smile:

Issue #510, opened 2020-04-30
[Feature request] match title variable for hostname format


topjor:
I would like to request a feature so the `match_title` matchconfig variable can be used inside the `get5_hostname_format`
e.g. `{MATCHTITLE}` would resolve to the match title

As an addition the default hostname format could be changed to `{MATCHTITLE}: {TEAM1} vs {TEAM2}`
or `Get5 {MATCHTITLE}: {TEAM1} vs {TEAM2}`

MikkelDK:
You can edit on the source-code and do "PR" =)

Drarig29:
@topjor I don't think we need to update the default value for `get5_hostname_format`.

> `match_title`: wrapper on the `mp_teammatchstat_txt` cvar, but can use {MAPNUMBER} and {MAXMAPS} as variables that get replaced with their integer values. In a BoX series, you probably don't want to set this since get5 automatically sets mp_teamscore cvars for the current series score, and take the place of the mp_teammatchstat cvars.

`match_title` has a complex behavior according to the paragraph above. Its default value is:

https://github.com/splewis/get5/blob/7318ef013d5fcb38cc05a97cd569dda407fdb0c5/scripting/get5/matchconfig.sp#L5

Maybe some people need/prefer a constant hostname for the entire match.

However, you would be able to use your own custom `get5_hostname_format` set to `{MATCHTITLE}: {TEAM1} vs {TEAM2}` with my pull request (PR).

topjor:
Ah okay, I never looked match_title, I am overwriting it with my panel anyway to the name of the match given in there and like to also use it now in the hostname format. for the default this would not make any sense. As for the PR thank you, you beat me to it. I would probably have done it myself in the comming week or in half a year when we need it again for our yearly lan party.

topjor:
Closing as #511 is merged

PR #511, opened 2020-04-30
Handle {MATCHTITLE} in get5_hostname_format

Drarig29:
Here's the only need change needed for this feature request : #510.

If you are OK with it, [this page](https://github.com/splewis/get5/wiki/Full-list-of-get5-cvars) of the Wiki will need to be updated.


topjor:
I would implement this in the format cvar string function, this way it is available to all strings that get formated with cvars like demo name etc.
https://github.com/splewis/get5/blob/2cb92339e4cce679b890d250d69c219aa03389f7/scripting/get5.sp#L1390

Drarig29:
Actually, yes good idea. I couldn't find this function when searching `"{MATCHID}"` and other substitute format values on GitHub...

topjor:
I always use my IDE to search in a project, I trust that way more and also makes looking at the code a lot easier

Drarig29:
Lol, same for me. But I don't know... this time I first researched a way to do it on GitHub and then I made the change on my IDE in my fork. Anyway 😅

splewis:
Thanks! I'll update the wiki.

PR #515, opened 2020-05-03
Use current map when skip veto and no maplist

sontrinh:
Use current map if in default list when skip_veto = 1 and no maplist (not always de_cache anymore)


splewis:
Seems like better behavior - thanks.

PR #517, opened 2020-05-04
Add KAST stat

redismx:
This PR adds the KAST statistic (number of rounds where a player either had **K**ill, **A**ssist, **S**urvival or **T**raded death). The KAST percentage can be easily derived by calculating KAST/Rounds_Played.

It's an useful stat used for calculating player ratings, more info here: https://www.hltv.org/news/20695/introducing-rating-20


drizer-tech:
Do not review/merge this yet. In this way I did it still isn't possible to calculate the KAST, as one can not tell whether the use survived *or* had a traded death *or* a kill, etc. This stat alone is useless.

redismx:
@splewis this is now ready for review.

Issue #529, opened 2020-05-22
Documentation on the JSON returned by get5_status


alcjzk:
Hey,

would it be possible to get proper documentation on the match state JSON returned by the `get5_status` command?

Drarig29:
Don't be afraid to look at the code 😊

Just search for `get5_status` in the code. You'll see it runs this method:
https://github.com/splewis/get5/blob/bb64ed8fe06278d1f5aadec1759d7af75728ec59/scripting/get5.sp#L1324

I don't know if you are speaking about the whole response of the `get5_status` command or only a part of it.

But here's some documentation of the whole response.

Here is the **static part**:

- `plugin_version`: Get5's version number.
- `commit`: Only here if `COMMIT_STRING` is defined (probably not your case).
- `gamestate`: A number representing the game's state. (using the [`Get5State` enum](https://github.com/splewis/get5/blob/51fe79d0da8131f7104e4a78551f4364f06be950/scripting/include/get5.inc#L6-L16))
  - **0** - `Get5State_None` = No setup has taken place.
  - **1** - `Get5State_PreVeto` = Warmup, waiting for the veto.
  - **2** - `Get5State_Veto` = Warmup, doing the veto.
  - **3** - `Get5State_Warmup` = Setup done, waiting for players to ready up.
  - **4** - `Get5State_KnifeRound` = In the knife round.
  - **5** - `Get5State_WaitingForKnifeRoundDecision` = Waiting for a !stay or !swap command after the knife.
  - **6** - `Get5State_GoingLive` = In the lo3 process.
  - **7** - `Get5State_Live` = The match is live.
  - **8** - `Get5State_PostGame` = Postgame screen + waiting for GOTV to finish broadcast.

 - `paused`: Is the match paused?
- `gamestate_string`: A human-readable gamestate (generated by [`GameStateString()`](https://github.com/splewis/get5/blob/bb64ed8fe06278d1f5aadec1759d7af75728ec59/scripting/get5/util.sp#L532-L553)) which is a translation of `gamestate`.
  - `"none"`
  - `"waiting for map veto"`
  - `"map veto"`
  - `"warmup"`
  - `"knife round"`
  - `"waiting for knife round decision"`
  - `"going live"`
  - `"live"`
  - `"postgame"`

Now, there may still be **one or two additional parts**.

If current game state is different than `Get5State_None`:
- `matchid`: The current match's id. You can set it in match configs, with the property which has the same name.
[...]

PR #534, opened 2020-05-28
Added get5_status schema

Drarig29:
Some days ago, I answered this issue (https://github.com/splewis/get5/issues/529) with a very detailed schema. I added it in the README, just as @splewis suggested. Because the README is intended for everybody, I didn't mention the code with links as I did in the issue.

I can't decide if I should keep the `commit` property in the static part... this is mostly for contributors I think.


jenrik (approved):

splewis:
I'd rather this be a subsection under "API for developers".

The convars section is more important for users to see, and this schema should be put below that.

splewis:
Thanks! Sorry for the delay in merging.

Issue #532, opened 2020-05-28
CSGO update broke get5_loadmatch_url


39ma:
get5_loadmatch_url ****
L 05/28/2020 - 17:49:07: [get5.smx] Failed to load match config: Failed to create HTTP GET request
L 05/28/2020 - 17:49:07: get5_event: {"matchid":"","params":{"reason":"Failed to create HTTP GET request"},"event":"match_config_load_fail"}
Failed to load match config.

ZenOfWolf:
It looks like a problem with Steamworks after the newest update. I think we just have to wait for steamworks to be updated.
There is already a post about it on the alliedmodders forum, [here](https://forums.alliedmods.net/showthread.php?t=229556&page=76#post2702746).

39ma:
Thank you. Yep, we have to wait till SteamWorks will update.
На 28.05.2020 18:23:27, Victor <notifications@github.com> писал:
It looks like a problem with Steamworks after the newest update. I think we just have to wait for steamworks to be updated.
There is already a post about it on the alliedmodders forum, here [https://forums.alliedmods.net/showthread.php?t=229556&amp;page=76#post2702746].

tuudik:
@splewis Can you add new SteamWorks to build scripts?
https://github.com/KyleSanderson/SteamWorks/releases
It works now with new SteamWorks release.

splewis:
It seems like the only change needed is for users to use a newer SteamWorks extension. 

There are no changes to the Steamworks extension's include file still: https://github.com/KyleSanderson/SteamWorks/blob/1.2.3b/Pawn/includes/SteamWorks.inc, so I'm not seeing why the plugin would be incompatible, or what this change does. Isn't the compiled identical still? Or am I missing something?

splewis:
(sorry, I should have commented on #537 instead, but oh well)

tuudik:
Sorry, I just thought having the updated SteamWorks during build is neccessary :) But I am very novice.

splewis:
Np, all the steamworks changes needed are runtime code changes within the SteamWorks extension itself  - the API the extension gives to the get5 plugin remains unaffected, hence not needing any plugin-compile-time changes.

tuudik:
So as I understand only SteamWorks had to be rebuilt due to Steam SDK change:
https://github.com/KyleSanderson/SteamWorks/commit/a5905873cc86bbf3f582fff5db4489c850b3c673
Correct?

[...]

PR #537, opened 2020-06-02
Add new SteamWorks for building

tuudik:
Should fix https://github.com/splewis/get5/issues/532

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,12 +22,14 @@ before_script:
     - SMPACKAGE="http://sourcemod.net/latest.php?os=linux&version=${SMVERSION}"
     - wget $SMPACKAGE
     - tar xfz $(basename $SMPACKAGE)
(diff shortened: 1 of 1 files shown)


splewis:
What's the rationale for this change? Has the steamworks API changed? Get5 doesn't package Steamworks alongside it on purpose, so I'm not sure what the practical impact of this would be.

tuudik:
The SteamWorks API has changed recently:
https://blog.counter-strike.net/index.php/2020/05/30199/

After I switched to the new SteamWorks.inc, things started to work for me again.
Otherwise the get5_loadmatch_url command didnt work.

There's also issue about it here: https://github.com/splewis/get5/issues/532

tuudik:
Closing pull request, since it's not neccessary to have updated SteamWorks during build, I guess :)

PR #538, opened 2020-06-03
Initial value for g_LogoBasePath

tuudik:
I found error in logs about not able to check for logo path, I think it's due to g_LogoBasePath not having any value initally. Not sure if this is correct approach.
The error itself
```
L 06/01/2020 - 15:30:31: [SM] Exception reported: Invalid path. An empty path string is not valid, use "." to refer to the current working directory.
L 06/01/2020 - 15:30:31: [SM] Blaming: get5_apistats.smx
L 06/01/2020 - 15:30:31: [SM] Call stack trace:
L 06/01/2020 - 15:30:31: [SM]   [0] DirExists
L 06/01/2020 - 15:30:31: [SM]   [1] Line 160, ./scripting/get5_apistats.sp::Get5_OnSeriesInit
L 06/01/2020 - 15:30:31: [SM]   [3] Call_Finish
...
```

@MoritzLoewenstein Can you validate my PR?

### DIFF
--- a/scripting/get5_apistats.sp
+++ b/scripting/get5_apistats.sp
@@ -61,6 +61,7 @@ public void OnPluginStart() {
   LogDebug("OnPluginStart version=%s", PLUGIN_VERSION);
   g_UseSVGCvar = CreateConVar("get5_use_svg", "1", "support svg team logos");
   HookConVarChange(g_UseSVGCvar, LogoBasePathChanged);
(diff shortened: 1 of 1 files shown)


MoritzLoewenstein:
Makes sense, it was an obvious bug (my bad) which should get patched.

splewis:
The code probably probably shouldn't be using a global variable, but this is fine - thanks for the fix.

PR #541, opened 2020-06-05
Fix #540

### LINKED ISSUE #540: Using get5_endmatch command twice to end a match during map veto
If making a feature request, you should delete all the pre-filled text here.
If reporting a bug, fill in the following sections:

### Expected behavior
When using get5_endmatch command during mapveto ,match will be ended after this command.  

### Actual behavior
When using get5_endmatch command during mapveto, Chat box will display message of both "An admin force ended the match." and "A captain left during the veto, pausing the veto",while actually the match is not ended.You have to use the get5_endmatch command again and this time the match is fully ended.  

Basically you have to use get5_endmatch twice to end a match during matpveto,I don't know if it was supposed to be like this or it is a bug.
  

### Steps to reproduce

- Plugin version: V0.7.2-dev Build #494 (2020-6-3 23:01:12)
- Sourcemod version: v1.11.0.6481  

- Steps to reproduce (please be specific):
1. Start a match through a match config,"skip_veto" set to 0
2. Use get5_endmatch during map veto.

yourmnbbn:
After studying the code I found that when you call get5_endmatch ,in the callback it will cancel the veto menu if it was displayed. So in the menu handler the MenuAction_Cancel will be passed to the parameter action.,and AbortVeto function will be called ,as a result game state will be changed from Get5State_None to Get5State_PreVeto.  

There is supposed to be a limit when calling AbortVeto function, to prevent miscalling the function in command get5_endmatch.

### DIFF
--- a/scripting/get5/mapveto.sp
+++ b/scripting/get5/mapveto.sp
@@ -286,7 +286,9 @@ public int MapVetoMenuHandler(Menu menu, MenuAction action, int param1, int para
     g_LastVetoTeam = team;
 
   } else if (action == MenuAction_Cancel) {
-    AbortVeto();
+    if(g_GameState == Get5State_Veto) {
+      AbortVeto();
+    }
 
   } else if (action == MenuAction_End) {
     delete menu;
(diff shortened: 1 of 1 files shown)


splewis:
Looks like a reasonable fix to me, though I can't say I've tested this. Thanks.

Issue #183, opened 2017-09-07
consider adding tar.gz builds to releases


janstuemmel:
* on a standard ubuntu server installation `unzip` isn't installed, as well as in docker image. 
* sourcemod and metamod are releasing `tar.gz` files

consider doing this for get5 as well.

splewis:
Sounds good to me. I'll do that for future releases.

PR #562, opened 2020-08-30
Build on GitHub actions

kroleg:
This is a basic config for building plugin using github actions. 
Not sure if you prefer to keep using travis amd smbuilder. Anyway i decided to share this. I can help with tweaking/finishing this if you are interested.
Almost solves #183 

Possible improvements:
- [x] use git submodules instead of downloading sm-json
- [ ] post a download link in PR / Create releases
- [x] also create .zip archive

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,8 +33,13 @@ jobs:
     - run: spcomp get5/scripting/get5_mysqlstats.sp -i steamworks -i $SM_JSON_INC_PATH -o $OUTPUT_SM_PATH/plugins/disabled/get5_mysqlstats.smx
 
     - run: tar -zcvf build.tar.gz output/
-
     - uses: actions/upload-artifact@v2
       with:
         name: build.tar.gz
         path: build.tar.gz
+
+    - run: zip -r build.zip output/
+    - uses: actions/upload-artifact@v2
+      with:
+        name: build.zip
+        path: build.zip
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,41 @@
+name: Build
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+
+    - uses: actions/checkout@v2
+      with:
+        repository: clugg/sm-json
+        path: 'json'
+
+    - uses: actions/checkout@v2
+      with:
+        path: 'get5'
+
+    - uses: rumblefrog/setup-sp@master
+      with:
+        version: '1.10.x'
+
+    - run: wget https://raw.githubusercontent.com/KyleSanderson/SteamWorks/master/Pawn/includes/SteamWorks.inc -P steamworks
+
+    - run: mkdir -p output/addons/sourcemod/plugins/disabled
+    - run: cp -R get5/cfg output
+    - run: cp -R get5/translations output/addons/sourcemod
+    - run: cp -R get5/configs output/addons/sourcemod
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,41 @@
+name: Build
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
jenrik, line 7: When the `ubuntu-latest` version is updated this has a chance of suddenly breaking the build. I would recommend using a specific version to avoid that scenario.
```suggestion
    runs-on: ubuntu-18.04
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    - uses: actions/checkout@v2
+      with:
+        path: 'get5'
+
+    - uses: rumblefrog/setup-sp@master
+      with:
+        version: '1.10.x'
+
+    - run: wget https://raw.githubusercontent.com/KyleSanderson/SteamWorks/master/Pawn/includes/SteamWorks.inc -P steamworks
```
jenrik, line 23: Pin the version used to avoid external changes breaking the building
```suggestion
    - uses: actions/checkout@v2
      with:
        repository: clugg/sm-json
        path: 'json'
        ref: 3a9e22b8b4056809c95b3329e9f779b9c0be34cc

    - uses: actions/checkout@v2
      with:
        path: 'get5'

    - uses: rumblefrog/setup-sp@50d937a464d62aed807037f61c1a5c9d59964f88
      with:
        version: '1.10.x'

    - run: wget https://raw.githubusercontent.com/KyleSanderson/SteamWorks/9fb35e2f96e1e539cb1c9c769b5dc5edf5d50b4b/Pawn/includes/SteamWorks.inc -P steamworks
```

jenrik (changes requested):
The version of external dependencies should be pinned to avoid having the build break suddenly. When switching to Github actions the old `.travis.yaml` should be deleted, there would be no need to have it build twice. I don't know how `spcomp` work so I can't comment on that part.
Other than the above this looks good, nice work!

.github/workflows/build.yml
kroleg: 1. pinned steamworks to 1.2.3c tag
2. sm-json now loaded from submodule

splewis:
>  post a download link in PR / Create releases
 also create .zip archive

Being able to support these does seem helpful. Presumably, using github actions could replace the Travis CI + my hosted jenkins instance for builds, right? 

If so - this seems like it would be preferable, since the configuration for Travis & jenkins are done separately, which makes it annoying to maintain.

> Not sure if you prefer to keep using travis amd smbuilder.

My preference would be to keep using it, but I wouldn't hold up this PR on that, it's definitely somewhat of an esoteric thing I just happen to use for development (and it primarily existed for packaging general server  config, it's not that necessary for just maintaining a plugin like this).

kroleg:
Just added zip file to artifacts.

> Presumably, using github actions could replace the Travis CI + my hosted jenkins instance for builds, right?

Yep. Once PR merged there will be a build for each commit made. And zip and tar.gz files will be attached to action as artifacts. You can see how it will look here https://github.com/kroleg/get5/actions/runs/240488265 

I don't have experience with releases on github (so i can't create a good github action for it) and releases aren't so frequent, so suggest that you make them manually for now. It will be pretty easy, just download builded archieves and reattach them to releases. But if you really need automatic releases i can try to setup it.

splewis:
Sorry for the delay - let's give this a shot now and I can get a better sense of how it works with the UI and such.

PR #563, opened 2020-08-30
Set loaded url from get5_loadmatch_url as loaded file in get5_status

jenrik:
This sets the path of the loaded match to the URL used in `get5_loadmatch_url` so that the URL shows up in `get5_status`

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -150,6 +150,7 @@ KeyValues g_StatsKv;
 
 ArrayList g_TeamScoresPerMap = null;
 char g_LoadedConfigFile[PLATFORM_MAX_PATH];
+char g_LoadedConfigUrl[PLATFORM_MAX_PATH];
 int g_VetoCaptains[MATCHTEAM_COUNT];        // Clients doing the map vetos.
 int g_TeamSeriesScores[MATCHTEAM_COUNT];    // Current number of maps won per-team.
 bool g_TeamReadyOverride[MATCHTEAM_COUNT];  // Whether a team has been voluntarily force readied.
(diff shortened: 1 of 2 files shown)


Drarig29:
@splewis did you see my review? I don't think all the changes were a good thing...

splewis:
I don't know what review you're referring to. I don't see any comments on this PR.

Drarig29:
It's weird...

Can you follow this link? 

https://github.com/splewis/get5/pull/563#discussion_r479808844

Drarig29:
Wow, I really need to stop messing with GitHub on my phone... Maybe it was still a draft. And don't mind about the missclick on revert.

Anyway, here is what I wanted to say in the review:

For me, the [line](https://github.com/splewis/get5/pull/563/commits/904cb203e9a80d9549b2892612f0b4dee400398d#diff-ccf8a04d07fd0839c118dfb2d59814a6R261) which copies the string `g_LoadedConfigUrl` inside `g_LoadedConfigFile` is a breaking change and misleading because it means putting a url instead of a file path...

So I think `g_LoadedConfigFile` should remain unchanged and keep containing a filename with the pattern `remote_config%d.json` because it's the actual config file being loaded.

But I still think adding `g_LoadedConfigUrl` is a good thing. And we should add it in the `get5_status` too by the way.

splewis:
Yeah, I considered whether it  would be better off as a separate value exposed in `get5_status`.

I figured it might not be worthwhile to have a separate value, given the `remote_config.json` value is hardcoded & therefore not very useful. But as you point out, the fact it has the `%d` does make it potentially a little more interesting.

I'd be okay with another PR to add a separate value in the `get5_status` schema (stored in a separate variable, as you note).

PR #564, opened 2020-09-06
Add loaded config url in get5_status

Drarig29:
This PR is an answer to [this comment](https://github.com/splewis/get5/pull/563#issuecomment-687699636).

- It removes the copy of the remote config **URL** in the **filepath** variable `g_LoadedConfigFile` and actually makes a distinction between `g_LoadedConfigFile` and `g_LoadedConfigUrl`.
- It adds `loaded_config_url` in the output of the `get5_status` command.

@jenrik beware that you wouldn't be able to have a URL inside `g_LoadedConfigFile` anymore if this PR is merged.

@splewis if you are willing to merge this PR, please take a look at #534 and merge it (if it suits you) first, so that I can add a last commit to update the README accordingly in this PR 😊

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -1343,6 +1343,7 @@ public Action Command_Status(int client, int args) {
   if (g_GameState != Get5State_None) {
     json.SetString("matchid", g_MatchID);
     json.SetString("loaded_config_file", g_LoadedConfigFile);
+    json.SetString("loaded_config_url", g_LoadedConfigUrl);
     json.SetInt("map_number", GetMapNumber());
 
     JSON_Object team1 = new JSON_Object();
--- a/scripting/get5/matchconfig.sp
+++ b/scripting/get5/matchconfig.sp
@@ -201,7 +201,6 @@ stock bool LoadMatchFromUrl(const char[] url, ArrayList paramNames = null,
                             ArrayList paramValues = null) {
   bool steamWorksAvaliable = LibraryExists("SteamWorks");
 
-
   char cleanedUrl[1024];
   strcopy(cleanedUrl, sizeof(cleanedUrl), url);
   ReplaceString(cleanedUrl, sizeof(cleanedUrl), "\"", "");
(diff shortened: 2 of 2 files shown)


jenrik (commented):
I think it's a poor idea to have both at the same time, it makes it ambiguous which one is in use. Having a match source path, and a source type field would be better in my opinion.

Drarig29:
Not ambiguous at all, since `g_LoadedConfigFile` will always contain `remote_config%d.json` when we get the config remotely with `get5_loadmatch_url` so the pattern is always the same.

Plus, the semantics are correct. And you know which **actual** config file to look into on your disk if you want to inspect something.

Whereas with a simple url, the config file served by the server could change at any time.

Personally in my use case, I have 16 servers running at the same time, with config files served on my web panel's api endpoints and their urls are static (one for each server). So config files are changing constantly.

My only way to know what config file has been loaded in each instance of a server is to assign each server a different ID which will be contained in the `%d` of the remote filename pattern.

Drarig29:
Maybe I should only include `g_LoadedConfigUrl` in the status output when the config file was remotely loaded.

Or set `g_LoadedConfigUrl` to a specific value when not used... What could this value be? Empty string?

And yes, I need to add a commit to reset `g_LoadedConfigUrl` somewhere. I haven't tested it yet.

jenrik:
can we at least add a source type field, because that would greatly helped with supporting libraries that can unpack data into predefined structures. These libraries generally can't unpack data without hand written deserializers when the data type depends on the content of a field or the presence of certain fields.

Drarig29:
Actually, maybe it's simpler to:

- abandon my PR 
- keep @jenrik's changes
- add `get5_server_id` in `get5_status` output

The only reason I wanted to keep `g_LoadedConfigFile` unchanged is because I could know exactly which config file I can inspect on disk.

Given that `g_LoadedConfigFile` always had the same pattern (it's hardcoded as a constant) and that only the server ID changes, I could do without it.

Issue #565, opened 2020-09-10
Getting connected player IP Address


EAbychkov:
It's more like a question rather than an issue, sorry, I don't know where else I can ask. 
I need to log **Player IP address** ones he connected to my server, is it possible to accomplished it somehow? 
I have this logging:
```03/28/2020 - 18:17:13: get5_event: {"matchid`":"c096526a-78e0-4b7e-a0bd-a7d502114cef","params":{"client":"oof<14><STEAM_1:1:115783675><>","map_number":0,"map_name":"de_vertigo"},"event":"player_connect"}```

Drarig29:
Get5 is not supposed to give you this information. But you can try the CS:GO command `status`, I know there is information about players... there might be the ip addresses too.

EAbychkov:
Thanks for your response.

TandelK:
@Drarig29 We can consider adding this function once and see the status ? I think it will be cool feature to add them in logs so Admins can have IPs of players on Online Tournament that they dont cheat giving their friends or someone else to play with . We can surely save them with log on option and save them in logs but storing in start of the match will give us better view. IPs can be changed in certain scenario if using VPN or something but most of the tournament host always ask to not play with VPNs and also some might have dynamic IPs but can be tracked properly.

Drarig29:
@TandelK do you mean you'd like to automatically run `status` when a match is started? It would be a good idea to have this is the logs...

EAbychkov:
@TandelK @Drarig29  This is exactly what I need, to have this info in the logs. Do you think it's possible to implement it in the next release?

Drarig29:
Yes, easily. The next release might arrive this month, as said [here](https://github.com/splewis/get5/issues/558#issuecomment-679776636).

splewis:
Adding the client IP to the `player_connect` event seems like it would the preferred solution here. Does that work for you @EAbychkov?

Drarig29:
Oh yeah! Very good idea!

EAbychkov:
> Adding the client IP to the `player_connect` event seems like it would the preferred solution here. Does that work for you @EAbychkov?

Yes, absolutely.

PR #567, opened 2020-09-12
Log native status on match going live

Drarig29:
This is answering the issue https://github.com/splewis/get5/issues/565.

It adds a console variable `get5_log_status`, disabled by default. If enabled, a `status` command will be run each time a match goes live.

This would be useful to check that a player doesn't give its account to a friend to play in his place.


Drarig29:
There is a better way of solving the issue.

PR #570, opened 2020-09-19
Series Cancel Event

### LINKED ISSUE #569: Match/Series Cancelled via Admin
While building [`get5_eventapi`](https://github.com/yannickgloster/get5_eventapi), a sourcemod plugin based off of the `get5_apistats` that sends a json request to a url with all the get5 events, I saw that there was no get5 if the match was forcefully cancelled by an admin.

yannickgloster:
Created an event that is called when there is a cancellation by an admin.
Closes #569

### DIFF
--- a/scripting/get5/eventlogger.sp
+++ b/scripting/get5/eventlogger.sp
@@ -181,9 +181,8 @@ public void EventLogger_SeriesEnd(MatchTeam winner, int t1score, int t2score) {
   EventLogger_EndEvent("series_end");
 }
 
-public void EventLogger_SeriesCancel(MatchTeam winner, int t1score, int t2score) {
+public void EventLogger_SeriesCancel(int t1score, int t2score) {
   EventLogger_StartEvent();
-  AddTeam(params, "team", winner);
   params.SetInt("team1_series_score", t1score);
   params.SetInt("team2_series_score", t2score);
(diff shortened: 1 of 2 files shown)


scripting/get5/eventlogger.sp
```diff
@@ -181,6 +181,14 @@ public void EventLogger_SeriesEnd(MatchTeam winner, int t1score, int t2score) {
   EventLogger_EndEvent("series_end");
 }
 
+public void EventLogger_SeriesCancel(MatchTeam winner, int t1score, int t2score) {
+  EventLogger_StartEvent();
+  AddTeam(params, "team", winner);
```
splewis, line 186: I don't think this field is worth adding if it's always TeamNone.

yannickgloster: I left it in there for competition but it also makes sense to remove it

yannickgloster: I had originally kept it as in the debug you kept it in:
```
LogDebug("Calling Get5_OnSeriesResult(winner=%d, team1_series_score=%d, team2_series_score=%d)",
           MatchTeam_TeamNone, g_TeamSeriesScores[MatchTeam_Team1],
           g_TeamSeriesScores[MatchTeam_Team2]);
```

Removed it in next commit based on your feedback

PR #572, opened 2020-09-30
added docker container for building and build instructions

Apfelwurm:
for an easy build process when developing on get5 i've written a little dockerfile and the instructions how to use it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+FROM debian:10
+MAINTAINER Alexander Volz (Alexander@volzit.de)
+
+ENV SMVERSION 1.10
+
+ENV _clean="rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*"
+ENV _apt_clean="eval apt-get clean && $_clean"
+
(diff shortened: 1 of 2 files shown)


Apfelwurm:
Typing error in readme...

Apfelwurm:
closed the wrong one... not my best day :D

splewis:
Thanks!

Drarig29:
@Apfelwurm It seems we need to build the image and run it each time a change is done in the code. Is it right?

Could there be a way of just running the same image after a change in the code?

Apfelwurm:
@Drarig29 currently thats the case because of the submodule initialisation, the sourcemod download and the SteamWorks.inc download. 
Indeed it would be much better to do it like you mentioned. But therefore i would have to write a script which would be executed when launching the container which does the cleanup. I try to schedule the changes for the next week to clean this up :)

Drarig29:
Nice! Thanks you 😃

Apfelwurm:
@Drarig29 https://github.com/splewis/get5/pull/587 :)

PR #573, opened 2020-09-30
readded https support for get5_loadmatch_url

Apfelwurm:
Removed forcing of http from matchconfig.sp and changed get5_loadmatch_url instructions to use quotation marks when spefifying a url with http:// or https:// in front


Apfelwurm:
Typing error in readme...

PR #574, opened 2020-09-30
readded https support for get5_loadmatch_url

Apfelwurm:
Removed forcing of http from matchconfig.sp and changed get5_loadmatch_url instructions to use quotation marks when spefifying a url with http:// or https:// in front

### DIFF
--- a/scripting/get5/matchconfig.sp
+++ b/scripting/get5/matchconfig.sp
@@ -208,9 +208,8 @@ stock bool LoadMatchFromUrl(const char[] url, ArrayList paramNames = null,
   strcopy(g_LoadedConfigUrl, sizeof(g_LoadedConfigUrl), cleanedUrl);
 
   if (steamWorksAvaliable) {
-    // Add the protocl strings. Only allow http since SteamWorks doesn't support http it seems?
-    ReplaceString(cleanedUrl, sizeof(cleanedUrl), "https://", "http://");
-    if (StrContains(cleanedUrl, "http://") == -1) {
+    // Add the protocol strings if missing (only http).
+    if (StrContains(cleanedUrl, "http://") == -1 && StrContains(cleanedUrl, "https://") == -1) {
       Format(cleanedUrl, sizeof(cleanedUrl), "http://%s", cleanedUrl);
     }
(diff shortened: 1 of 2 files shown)


splewis:
I'm guessing you tested this and verified it works? I had briefly tested https support, but didn't get it to work. If adding quotation is all it took, that's great to know.

Apfelwurm:
> I'm guessing you tested this and verified it works? I had briefly tested https support, but didn't get it to work. If adding quotation is all it took, that's great to know.

Yes, we use it since a few days now and it seems to work perfectly :) it looks like SM struggles with the two slashes in the console, if you don't use quotation. even with http this was the case.

splewis:
Great - thanks for confirming.

scripting/get5/matchconfig.sp
```diff
@@ -208,9 +208,8 @@ stock bool LoadMatchFromUrl(const char[] url, ArrayList paramNames = null,
   strcopy(g_LoadedConfigUrl, sizeof(g_LoadedConfigUrl), cleanedUrl);
 
   if (steamWorksAvaliable) {
-    // Add the protocl strings. Only allow http since SteamWorks doesn't support http it seems?
-    ReplaceString(cleanedUrl, sizeof(cleanedUrl), "https://", "http://");
-    if (StrContains(cleanedUrl, "http://") == -1) {
+    // Add the protocol strings if missing (only http).
+    if (StrContains(cleanedUrl, "http://") == -1 && StrContains(cleanedUrl, "https://") == -1) {
```
splewis, line 212: This expression seems like it will never be true anymore, right? 

To be true it would have to contain both `http://` and `https://`, which seems largely impossible for any sane url.

Drarig29: `StrContains()` returns `-1` when not found or the index of the match otherwise. Therefore, `http://` will be added if there is neither http nor https 😁

Apfelwurm: @Drarig29 is right :) We have also tested this part of the codechange.

splewis: Ah, yes, I didn't read carefully enough - I hate that SM's `StrContains` doesn't return a bool. Thanks

PR #580, opened 2020-10-08
Add information in get5_listbackups

### LINKED ISSUE #578: Add details in get5_listbackups
Currently, the command `get5_listbackups` just lists the filename of all backup files but there is no more information.

The filename of a backup file is `get5_backup_match%s_map%d_round%d.cfg` or `get5_backup_match%s_map%d_prelive.cfg`, so we only have the match id, the map number and the round number. But it's pretty good like this, so I think it should stay as is.

The problem is that in a LAN with 16 servers (my case), it's quite hard to determine which backup file to choose when a team comes to your desk to ask to load the backup for "the round with a score of 14-12 in the match Team A vs. Team B".

On our private fork of get5, we changed `Command_ListBackups()` to output the following information:
- If there is no Valve backup in the file: `filename date team1_name team2_name`
- Otherwise: `filename date team1_name team2_name map_name team1_score team2_score`

It's done by reading wanted information in each backup file.

If we don't want to change the behavior of the existing command, we could add an optional argument to it (just like `ls` has `ls -l`) or create another command.

I already have the code so I can make a PR.

Drarig29:
Closes #578 

This PR adds the function `GetBackupInfo()` which reads a backup file to retrieve more information about it.

The possible output of the function is the following:
- If there is no Valve backup in the file: `filename date team1_name team2_name`
- Otherwise: `filename date team1_name team2_name map_name team1_score team2_score`
- Returns `false` if it fails (read failure or at least one empty team name, which would break the format) and logs an error

If this function fails, the fallback line for a backup will be the old style line (only `filename`).

### DIFF
--- a/scripting/get5/backups.sp
+++ b/scripting/get5/backups.sp
@@ -26,6 +26,7 @@ public Action Command_ListBackups(int client, int args) {
     ReplyToCommand(client, "The backup system is disabled");
     return Plugin_Handled;
   }
+  
   char matchID[MATCH_ID_LENGTH];
   if (args >= 1) {
     GetCmdArg(1, matchID, sizeof(matchID));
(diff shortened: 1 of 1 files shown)


scripting/get5/backups.sp
```diff
@@ -50,6 +57,80 @@ public Action Command_ListBackups(int client, int args) {
...
+  }
+
+  // Try entering Valve's backup section (it doesn't always exist).
+  if (!kv.JumpToKey("valve_backup")) {
+    Format(info, maxlength, "%s %s %s %s", path, timestamp, team1Name, team2Name);
+    delete kv;
+    return true;
+  }
+
+  char map[15];
```
splewis, line 100: Let's use 128 here.

scripting/get5/backups.sp
```diff
@@ -39,9 +40,15 @@ public Action Command_ListBackups(int client, int args) {
   DirectoryListing files = OpenDirectory(".");
   if (files != null) {
     char path[PLATFORM_MAX_PATH];
+    char backupInfo[150];
```
splewis, line 43: I think this is worth making longer as well, perhaps 256.

Drarig29:
Would you mind adding a `hacktoberfest-accepted` label to this PR? It's just so that it adds in my PR counter.

It wouldn't make your repo opt in [the event] (https://hacktoberfest.digitalocean.com/) and you would be able to remove the label next month.

splewis:
Done.

PR #587, opened 2020-10-29
changed docker building process to one time docker build

Apfelwurm:
Thanks to @Drarig29 's [comment](https://github.com/splewis/get5/pull/572#issuecomment-718088614) on the docker build process i have updated the Dockerfile so it is CI compatible.
You can now build the Docker container only one time and use the run command to build get5 as often you need without the need of a rebuild if you change code on get5.


Apfelwurm:
@splewis ? :)

PR #590, opened 2020-11-20
Added command to add the last kicked player

### LINKED ISSUE #589: Add a command to add the last player that was kicked
As an admin, sometimes I need to add players with `get5_addplayer`.

I can run RCON commands from my webpanel but I need to look at the server logs to find the last player that was kicked with reason "You are not a player in this match" and copy paste his SteamID and his name.

It would be cool to have a command `get5_addkickedplayer <team1|team2|spec>`.

I don't think a command `get5_lastkickedplayer` would be useful because we would still need to copy paste the result and use it with `get5_addplayer`.

I think this should be implemented in get5 because it's the only plugin to know *why* the player is being kicked (not a player in this match) and get5 could easily remember the SteamID and the name of the last kicked player.

I would make a PR on this.

Drarig29:
Closes #589 

The last kicked player is saved in `g_LastKickedPlayerAuth`. (By the way, I don't know why, but I couldn't use `AUTH_LENGTH` for the size of this char array, it wasn't compiling...)

All client kicks with reason "YourAreNotAPlayerInfoMessage" are now saving the player auth in the global variable.

If for whatever reason, an admin wants to undo the effect of this command (remove the added player), he can do the following :

- Fake a new `get5_addkickedplayer` call, with any team as argument (`team1|team2|spec`)
- This command will fail because the player is already added, but **it will print the steamid** that he needs in the error message
- Use `get5_removeplayer` with the aforementioned steamid (copy-paste)

I don't reset `g_LastKickedPlayerAuth` after the command to enable the latter scenario. This variable will be overwritten when another player gets kicked.

The command can't be used as long as no player is kicked.

Maybe a command `get5_removekickedplayer` would be a good thing to add to complete the pair of commands...

I didn't think about that before. This would save admins from having to use the workaround scenario 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ Some client commands are available also for admin usage. For example, sm_pause a
 - ``get5_creatematch``: creates a Bo1 match with the current players on the server on the current map
 - ``get5_scrim``: creates a Bo1 match with the using settings from ``configs/get5/scrim_template.cfg``
 - ``get5_addplayer``: adds a steamid to a team (any format for steamid)
+- ``get5_addkickedplayer``: adds the last kicked steamid to a team
 - ``get5_removeplayer``: removes a steamid from all teams (any format for steamid)
 - ``get5_forceready``: marks all teams as ready
 - ``get5_dumpstats``: dumps current match stats to a file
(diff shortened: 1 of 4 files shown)


splewis:
`get5_removekickedplayer`  sounds good to me as well, but I won't hold up this PR on it (we can do it separate). Thanks.

PR #608, opened 2020-12-23
MVP Stats + MySQL best practices and UTF8

nickdnk:
Hello

I plan on starting to collect stats via MySQL and I came across some minor problems with the current integration.

1. Use of `SELECT LAST_INSERT_ID()` to retrieve the latest insert ID is unnecessary given the SourceMod driver. You can retrieve the latest auto-increment ID used from the `InsertId` property of the result-set returned after performing an insert. I've adjusted this and tested that it works as expected. It gets rid of the transaction altogether.

3. Use of `REPLACE INTO ...` instead of `INSERT INTO (...) ON DUPLICATE KEY UPDATE` is a bad idea. Not only is it much, much slower, but it actually deletes and re-inserts the row(s), which can cause problems for foreign keys if the rows are referenced elsewhere. I've changed this behavior when updating player stats.

3. UTF8 in MySQL is not really UTF8. It's only 3 bytes. The "real" UTF8 charset is called `utf8mb4`, which is also the default in MySQL 8+. The "old" `utf8` won't support emojis and other 4-byte characters. I adjusted this in the `import_stats.sql` script. See https://stackoverflow.com/questions/30074492/what-is-the-difference-between-utf8mb4-and-utf8-charsets-in-mysql.

4. There is no index defined on the `steamid64` column in the player stats table. This is highly relevant as it would be used when querying stats for any specific player. I've added this.

5. There are no foreign keys defined on tables that reference each other. I've added these, so the players and maps tables reference the match table.

### DIFF
--- a/misc/import_stats.sql
+++ b/misc/import_stats.sql
@@ -65,6 +65,7 @@ CREATE TABLE `get5_stats_players`
     `tradekill`          smallint(5) unsigned NOT NULL,
     `kast`               smallint(5) unsigned NOT NULL,
     `contribution_score` smallint(5) unsigned NOT NULL,
+    `mvp`                smallint(5) unsigned NOT NULL,
     PRIMARY KEY (`matchid`, `mapnumber`, `steamid64`),
     KEY `steamid64` (`steamid64`),
     CONSTRAINT `get5_stats_players_matchid` FOREIGN KEY (`matchid`) REFERENCES `get5_stats_matches` (`matchid`)
--- a/misc/import_stats.sql
+++ b/misc/import_stats.sql
@@ -1,112 +1,72 @@
---
--- Table structure for table `get5_stats_maps`
---
-
-DROP TABLE IF EXISTS `get5_stats_maps`;
-/*!40101 SET @saved_cs_client     = @@character_set_client */;
-/*!40101 SET character_set_client = utf8 */;
-CREATE TABLE `get5_stats_maps` (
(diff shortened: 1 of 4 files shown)


nickdnk:
I also added support for recording the MVP column of the stats using the `round_mvp` event.

splewis:
I'm okay with this, but I'd like to mention I don't really want to support anything more complicated on the sql front - I'd rather people build their own stats collections on top of the core get5 API only (i.e. make your own version of the get5_mysqlstats or get5_apistats plugins). 

That said, these changes seem okay. I'm not concerned with compatibility breakage for the aforementioned reason (only the core  get5 plugin is intended to support any level of backwards compatibility).

ErcouldnT:
If someone checks compatibility of get5_mysqlstats and get5_apistats plugins, tell me please. I don't know should I download it or not?

nickdnk:
@ErcouldnT The changes made in this particular PR would require that you add a column called "mvp" to the players table. See the sample SQL script (which was also updated in this PR) and make sure your tables match.

ErcouldnT:
Excuse me, how can I create new database with new column?

nickdnk:
@ErcouldnT If you want to "create a new database" you can just drop the entire get5 database and run the script again. But that way you'll lose all your data. If you just want to add the column, run this command on the get5 database.

`ALTER TABLE get5_stats_players ADD COLUMN mvp SMALLINT(5) UNSIGNED NOT NULL AFTER contribution_score`

PR #609, opened 2020-12-24
Added knife kills and utility damage tracking

nickdnk:
Hey again!

I should have added this to my previous PR, but you beat me to merge by an hour. Sorry!

I added utility damage and knife kill tracking as well. The UD is as close to the in-game scoreboard as I can get it, with the only difference being that this will include damage dealt from grenades physically hitting you, while the in-game scoreboard only includes damage dealt from the actual explosion/fire (typically 2-3 damage difference). Aside from that I made sure that it's counted exactly the same and excludes damage dealt to self or your own team. The existing `damage` stat also incorrectly included any kind of damage inflicted to yourself or your team, so I have corrected that as well.

RE: The comment you made on the previous PR: I understand the desire to avoid cluttering this more than necessary, but I also **really** don't want to maintain my own version of this and do manual merges and rebuilds for such few changes, as it's 99% of what I need. I'd very much appreciate it if you could merge this as well.

Merry Christmas :)

Edit: I noticed that you had a comment about `Format()` breaking at 32 arguments, but with this now at 33 it still works fine. Maybe this has been fixed with an update of SourceMod. I have not removed the comment though.

### DIFF
--- a/scripting/get5/readysystem.sp
+++ b/scripting/get5/readysystem.sp
@@ -67,6 +67,10 @@ public bool IsTeamReady(MatchTeam team) {
   int playerCount = GetTeamPlayerCount(team);
   int readyCount = GetTeamReadyCount(team);
 
+  if (team == MatchTeam_TeamSpec && minReady == 0) {
+    return true;
+  }
+
   if (playerCount == readyCount && playerCount >= minPlayers) {
     return true;
   }


scripting/get5/stats.sp
```diff
@@ -227,8 +227,13 @@ public Action Stats_PlayerDeathEvent(Event event, const char[] name, bool dontBr
       IncrementPlayerStat(attacker, STAT_KILLS);
       g_PlayerRoundKillOrAssistOrTradedDeath[attacker] = true;
 
-      if (headshot)
+      if (headshot) {
         IncrementPlayerStat(attacker, STAT_HEADSHOT_KILLS);
+      }
+
+      if (StrEqual(weapon, "knife") || StrEqual(weapon, "knife_t")) {
```
nickdnk, line 234: The force-pushed was to add `knife_t`, as I found out during testing that the knife has a different value on the T side. GG Valve.

splewis:
Looks like the max for calling a function is 127: https://github.com/alliedmodders/sourcepawn/blob/e8d7b67b0665baf1faf6d4985c1e92ac1163e43f/include/sp_vm_types.h#L49-L50

This got tweaked in https://github.com/alliedmodders/sourcepawn/commit/d610630aa61aed4fe18d876cc1ba0ae293cc3fae. 

Since it's not a function we're declaring - only calling - I'm okay with this. Can you update the comment to mention the 127 args limit, and say "see SP_MAX_CALL_ARGUMENTS in sourcepawn."

nickdnk:
Sure thing. Do you want me to add the changes I made to get5_apistats? I made no changes to that plugin at all, but it's quite easy to add mvp, knife kills and utility damage to that as well.

scripting/get5/stats.sp
```diff
@@ -227,8 +227,16 @@ public Action Stats_PlayerDeathEvent(Event event, const char[] name, bool dontBr
...
 
-      if (headshot)
+      if (headshot) {
         IncrementPlayerStat(attacker, STAT_HEADSHOT_KILLS);
+      }
+
+      // Because custom knives all have different names, we simply search the weapon string.
+      // StrContains returns the index and -1 if the string was not found.
+      // As all knives must *start with* 'knife', we can be sure that this must return 0 for all knives.
+      if (StrContains(weapon, "knife", true) == 0) {
```
nickdnk, line 237: So it turns out that all custom knives have different string values as well, so I replaced this with a `StrContains()`. I am going to assume that all knives start with `knife`, as I have limited knives available and that was the case for all of them. This does makes sense as it indicates to the game which icon to show in the kill-feed, I just never thought of that.

Issue #584, opened 2020-10-15
Add serverid & ip:port into mysqlstats data


RuneNyhuus:
Hey

Is it possible to somehow make it type the IP / Hostname into the Mysql_stats?
Im runninng multiple servers, but i have no way to identify what server is making the stats.
Would be nice if it could write IP or Hostname into the table in the mysql.

I know many would say make different databases for each server, and yes i could do that.
But im deploying alot of servers all the time, so if i could deploy all with the same sql settings.

I see this options in the get5.cfg
// Integer that identifies your server. This is used in temp files to prevent collisions.
// -
// Default: "0"
get5_server_id "0"

But it do not write that server_id into the sql

Or maybe there is another way for this?

emil6156:
+1..
Im struggling with exact the same thing.

the-jomper:
Me to.

VictorABoye:
I've been looking into this, though I haven't used sourcepawn before. But I have a question for you guys before I continue: 
Would it be preferred to have IP, Hostname or the get5_server_id?

Also just a note, it looks like the max amount of arguments have been reached for the get5_stats_players table, atleast in the sql plugin file. I think it would be better if someone more adept at SourcePawn would look at that. 
Therefore, I would only be able to add identifiers to get5_stats_maps and get5_stats_matches. Which I am looking into adding.

RuneNyhuus:
> I've been looking into this, though I haven't used sourcepawn before. But I have a question for you guys before I continue:
> Would it be preferred to have IP, Hostname or the get5_server_id?
> 
> Also just a note, it looks like the max amount of arguments have been reached for the get5_stats_players table, atleast in the sql plugin file. I think it would be better if someone more adept at SourcePawn would look at that.
> Therefore, I would only be able to add identifiers to get5_stats_maps and get5_stats_matches. Which I am looking into adding.

IP or get5_server_id would be preferred.
If its possible to get get5_server_id into get5_stats_maps or get5_stats_matches, it would be totally fine for me.

splewis:
I'm fine with adding both `get5_server_id` and a 'host:ip' into the `get5_stats_matches` table. PRs welcome. I don't think it's worthwhile to add it to the other tables, so the max-arg limit shouldn't be an issue.

RuneNyhuus:
Hey @splewis can you write here when the feature is added... Looking forward to it!! 👍

nickdnk:
[...]

PR #613, opened 2021-01-03
Store server ID in match table

nickdnk:
Hello

As discussed in https://github.com/splewis/get5/issues/584.

But there's some code here I don't understand. This PR needs some help as I'm not 100% sure this is right. Can you explain why this code has reversed get/set logic and always returns 0? That makes no sense at all to me. 

https://github.com/splewis/get5/blob/013519d35135a750c157e9f5919232a538873c79/scripting/get5/natives.sp#L185-L195

I ask because I used the same pattern for this PR. I did it in the more conventional "return the integer"-way though, which may be wrong.

CC @RuneNyhuus

### DIFF
--- a/misc/import_stats.sql
+++ b/misc/import_stats.sql
@@ -13,6 +13,7 @@ CREATE TABLE `get5_stats_matches`
     `team1_score` smallint(5) unsigned NOT NULL DEFAULT '0',
     `team2_name`  varchar(64)          NOT NULL DEFAULT '',
     `team2_score` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `server_id`   int(10) unsigned     NOT NULL DEFAULT '0',
     PRIMARY KEY (`matchid`)
 ) ENGINE = InnoDB
   DEFAULT CHARSET = utf8mb4;
--- a/misc/import_stats.sql
+++ b/misc/import_stats.sql
@@ -14,14 +14,15 @@ CREATE TABLE `get5_stats_matches`
     `team2_name`  varchar(64)          NOT NULL DEFAULT '',
     `team2_score` smallint(5) unsigned NOT NULL DEFAULT '0',
     `server_id`   int(10) unsigned     NOT NULL DEFAULT '0',
-    PRIMARY KEY (`matchid`)
+    PRIMARY KEY (`matchid`),
+    KEY `server_id` (`server_id`)
 ) ENGINE = InnoDB
   DEFAULT CHARSET = utf8mb4;
 
 CREATE TABLE `get5_stats_maps`
 (
     `matchid`     int(10) unsigned     NOT NULL,
-    `mapnumber`   smallint(5) unsigned NOT NULL,
+    `mapnumber`   tinyint(3) unsigned  NOT NULL,
     `start_time`  datetime             NOT NULL,
     `end_time`    datetime             NULL     DEFAULT NULL,
     `winner`      varchar(16)          NOT NULL DEFAULT '',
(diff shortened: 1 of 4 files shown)


Drarig29:
> Can you explain why this code has reversed get/set logic and always returns 0? That makes no sense at all to me.

That's how natives work in SourceMod. A native doesn't use the return value but uses functions to set and get "native" parameters. In this context, `return 0` means it worked well.

For example, with `Native_GetMatchID()`, you need to "return" a string, so you "set" the native string which will be returned with the [`SetNativeString()`](https://sm.alliedmods.net/new-api/functions/SetNativeString) function.

And in `Native_SetMatchID()`, you use [`GetNativeString()`](https://sm.alliedmods.net/new-api/functions/GetNativeString), which gets the string parameter passed to the native and directly copies the value in the `g_MatchId` variable

misc/import_stats.sql
```diff
@@ -13,6 +13,7 @@ CREATE TABLE `get5_stats_matches`
     `team1_score` smallint(5) unsigned NOT NULL DEFAULT '0',
     `team2_name`  varchar(64)          NOT NULL DEFAULT '',
     `team2_score` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `server_id`   smallint(5) unsigned NOT NULL DEFAULT '0',
```
Drarig29, line 16: I would have used a `int`, and maybe `int(10) unsigned` as for the matchid.

Drarig29 (changes requested):
The more conventional "return the integer"-way is fine imo, it's already the case for `Get5_MatchTeamToCSTeam()` and `Get5_IncreasePlayerStat()`

misc/import_stats.sql
nickdnk: You're right. smallint is probably too small. I'll change it.

Drarig29 (approved):

nickdnk:
For anyone using the MySQL currently, you must run this command on the get5 database when this gets merged:

`ALTER TABLE get5_stats_matches ADD COLUMN server_id INT(10) UNSIGNED NOT NULL DEFAULT '0' AFTER team2_score`

nickdnk:
Hey @Drarig29

I made a few adjustments to the SQL script while we're at it.

1. There's no reason `mapnumber` cannot be tinyint unsigned. I don't see any way a match would ever have more than 255 maps.
2. I changed `steamid64` to bigint, as I noticed it comes from the server and is always in the STEAM64 format. There's no reason to store that as varchar. Note that the 21 length is an arbitrary limit and those have been removed for all integer types in MySQL 8 anyway.
3. I added an index to the `server_id` column we just added in this PR as I forgot that originally.

All of the above is backwards-compatible.

Drarig29:
I'm OK with 1. and 3. but maybe @splewis should take a look into 2. I'm OK with 2. too but I might forget something.

nickdnk:
Sure thing. The only reason I didn't change 2. in my previous contribution to the SQL script was because I thought you had made it varchar to support different types of Steam IDs, just like the matchconfig file does, but after taking a closer look at the code it seems this is not the case. I'm not 100% sure though.

A SteamID as varchar in utf8mb4 is ~69 (17 characters x4 + 1 for length) bytes, but only 8 as bigint. When used in a primary key this means 61 bytes are wasted per row - plus an additional 61 per row for each column you define a secondary index on. Quite significant if you store a lot of matches: Say you add two secondary indexes. That's 207 bytes per row (69x3) out the window. A thousand rounds (that's only 40 games at an average of 25 rounds per game) played with 10 players (1,000 x 10; x10,000) and it's 2,070 KB or ~2.02 MB of disk use and memory allocated for no reason at all. As a DBA that hurts my soul 😭😄

RuneNyhuus:
So i have made the changes to the SQL database that @nickdnk posted.
What files do i need to change?
Sorry im new to this Github :)

Drarig29:
Don't do it yet, the PR still open and more changes could be added.

RuneNyhuus:
> Don't do it yet, the PR is still open and more changes could be added.

Ok thank you for reply @Drarig29 :)

Issue #412, opened 2019-06-18
Include an Get5_OnMatchPause and Get5_OnMatchUnpause event


PhlexPlexico:
Hi!

I'd like to make a suggestion to include a new forward for whenever a match is paused or unpaused within get5. The way I've seen it justified is that if there is a way to detect pauses, a user could create their own plugin to track the forward and store the values/provide feedback on matches that are paused to their own services (i.e. API responses). It could also be led to more data collection on a service that uses get5 to see how many average pauses are taken per match  over a period of games, as an example.

If there is not a way of tracking this already, would this be a good suggestion to look into?

Thanks!

splewis:
There's an existing PR that has unresolved problems: https://github.com/splewis/get5/pull/386

vassilidev:
sm_unpause

PR #626, opened 2021-01-22
Pause/Unpause events and forwards

PhlexPlexico:
Hi!

Per the issue #412, this PR creates some forwards and events for pausing and un-pausing in game, so that they can be used in other applications/plugins as well (such as a web-panel notifying if a match is paused).

I saw that PR #386 hasn't been acted upon since I last looked, and I figured I would take a stab at the fixed timeout issue mentioned, as well as provide some context for a pause.

The pause events also take in which team is making the call, and a reason, either tactical or technical, this is to provide some context and help with a forward to determine why a pause was called.

I tested a few of the different possibilities for pausing, including fixed timer and tech pauses, and I think the events successfully pause/unpause under every event.

### DIFF
--- a/scripting/get5/eventlogger.sp
+++ b/scripting/get5/eventlogger.sp
@@ -52,6 +52,12 @@ static void AddTeam(JSON_Object params, const char[] key, MatchTeam team) {
   params.SetString(key, value);
 }
 
+static void AddPause(JSON_Object params, const char[] key, PauseType pause) {
+  char value[16];
+  GetPauseType(pause, value, sizeof(value));
+  params.SetString(key, value);
+}
+
 static void AddCSTeam(JSON_Object params, const char[] key, int team) {
   char value[16];
   CSTeamString(team, value, sizeof(value));
@@ -269,4 +275,19 @@ public void EventLogger_TeamUnready(MatchTeam team) {
   AddTeam(params, "team", team);
 
   EventLogger_EndEvent("team_unready");
+}
+
+public void EventLogger_PauseCommand(MatchTeam team, PauseType pauseReason) {
+  EventLogger_StartEvent();
+  AddMapData(params);
+  AddTeam(params, "request_team", team);
+  AddPause(params, "pause_reason", pauseReason);
+  EventLogger_EndEvent("pause_command");
+}
+
+public void EventLogger_UnpauseCommand(MatchTeam team) { 
+  EventLogger_StartEvent();
+  AddMapData(params);
+  AddTeam(params, "request_team", team);
+  EventLogger_EndEvent("unpause_command");
 }
--- a/scripting/get5/pausing.sp
+++ b/scripting/get5/pausing.sp
@@ -11,11 +11,24 @@ public Action Command_TechPause(int client, int args) {
 
   if (client == 0) {
     Pause();
+    EventLogger_PauseCommand(MatchTeam_TeamNone, PauseType_Tech);
+    LogDebug("Calling Get5_OnMatchPaused(team=%d, pauseReason=%d)", MatchTeam_TeamNone, PauseType_Tech);
+    Call_StartForward(g_OnMatchPaused);
+    Call_PushCell(MatchTeam_TeamNone);
+    Call_PushCell(PauseType_Tech);
+    Call_Finish();
     Get5_MessageToAll("%t", "AdminForceTechPauseInfoMessage");
     return Plugin_Handled;
   }
 
+  MatchTeam team = GetClientMatchTeam(client);
   Pause();
+  EventLogger_PauseCommand(team, PauseType_Tech);
+  LogDebug("Calling Get5_OnMatchPaused(team=%d, pauseReason=%d)", team, PauseType_Tech);
+  Call_StartForward(g_OnMatchPaused);
+  Call_PushCell(team);
+  Call_PushCell(PauseType_Tech);
+  Call_Finish();
   Get5_MessageToAll("%t", "MatchTechPausedByTeamMessage", client);
 
   return Plugin_Handled;
@@ -32,6 +45,12 @@ public Action Command_Pause(int client, int args) {
     g_InExtendedPause = true;
 
     Pause();
+    EventLogger_PauseCommand(MatchTeam_TeamNone, PauseType_Tactical);
+    LogDebug("Calling Get5_OnMatchPaused(team=%d, pauseReason=%d)", MatchTeam_TeamNone, PauseType_Tactical);
+    Call_StartForward(g_OnMatchPaused);
+    Call_PushCell(MatchTeam_TeamNone);
+    Call_PushCell(PauseType_Tactical);
+    Call_Finish();
     Get5_MessageToAll("%t", "AdminForcePauseInfoMessage");
     return Plugin_Handled;
   }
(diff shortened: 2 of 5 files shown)


scripting/get5/eventlogger.sp
```diff
@@ -78,10 +84,8 @@ public void EventLogger_SeriesStart() {
 
 public void EventLogger_MapVetoed(MatchTeam team, const char[] map) {
   EventLogger_StartEvent();
-
```
Drarig29, line 81: Try not to change code in places you are not doing anything important. And maybe follow the existant style.

Like, keep the same style with blank lines in the event logger functions for `Pause` and `Unpause` too. 😊

scripting/get5/eventlogger.sp
```diff
@@ -249,17 +252,28 @@ public void EventLogger_PlayerDisconnect(int client) {
...
+
+public void EventLogger_PauseCommand(MatchTeam team, PauseType pauseReason) {
+  EventLogger_StartEvent();
+  AddMapData(params);
+  AddTeam(params, "request_team", team);
+  AddPause(params, "pause_reason", pauseReason);
+  EventLogger_EndEvent("pause_command");
+}
+
+public void EventLogger_Unpause(MatchTeam team) {
```
Drarig29, line 274: The Command suffix is missing in `EventLogger_UnpauseCommand`.

scripting/get5/pausing.sp
```diff
@@ -59,13 +77,24 @@ public Action Command_Pause(int client, int args) {
 
   // If the pause will need explicit resuming, we will create a timer to poll the pause status.
   bool need_resume = Pause(g_FixedPauseTimeCvar.IntValue, MatchTeamToCSTeam(team));
+  EventLogger_PauseCommand(team, PauseType_Tac);
+  LogDebug("Calling Get5_OnMatchPaused(team=%d, pauseReason=%d)", team, PauseType_Tac);
+  Call_StartForward(g_OnMatchPaused);
+  Call_PushCell(team);
+  Call_PushCell(PauseType_Tac);
+  Call_Finish();
```
Drarig29, line 85: A blank line after this? 😅 (sorry, nitpicking)

scripting/get5/pausing.sp
```diff
@@ -11,11 +11,23 @@ public Action Command_TechPause(int client, int args) {
...
+    EventLogger_PauseCommand(MatchTeam_TeamNone, PauseType_Tech);
+    LogDebug("Calling Get5_OnMatchPaused(team=%d, pauseReason=%d)", MatchTeam_TeamNone, PauseType_Tech);
+    Call_StartForward(g_OnMatchPaused);
+    Call_PushCell(MatchTeam_TeamNone);
+    Call_PushCell(PauseType_Tech);
+    Call_Finish();
     Get5_MessageToAll("%t", "AdminForceTechPauseInfoMessage");
     return Plugin_Handled;
   }
-
```
Drarig29, line 17: Do not remove this line 😊

scripting/get5/pausing.sp
```diff
@@ -90,6 +119,45 @@ public Action Command_Pause(int client, int args) {
...
+    g_PauseTimerHandle = INVALID_HANDLE;
+    return Plugin_Stop;
+  }
+
+  if (!InFreezeTime()) {
+    // Someone can call pause during a round and will set this timer.
+    // Keep running timer until we are paused.
+    return Plugin_Continue;
+  } else {
+    if (g_PauseTimeUsed <= 0) {
```
Drarig29, line 142: This if instruction is badly indented.

scripting/get5/pausing.sp
```diff
@@ -59,13 +77,24 @@ public Action Command_Pause(int client, int args) {
...
+  Call_Finish();
   if (IsPlayer(client)) {
     Get5_MessageToAll("%t", "MatchPausedByTeamMessage", client);
   }
 
   if (IsPlayerTeam(team)) {
     if (need_resume) {
+      g_PauseTimeUsed = g_PauseTimeUsed + g_FixedPauseTimeCvar.IntValue - 1;
       CreateTimer(1.0, Timer_PauseTimeCheck, team, TIMER_REPEAT | TIMER_FLAG_NO_MAPCHANGE);
+      // Keep track of timer, since we don't want several timers created for one pause checking instance.
```
Drarig29, line 94: Are you sure you need to keep track of it? This would be the only case of that in the project.

Drarig29 (changes requested):
I have some suggestions on what to change.

And maybe you should explain more the changes you made, especially the `Timer_UnpauseEventCheck()`.

Basically, tell that:

- You cancel the timer if not pauseable
- You cancel the timer if the pause time is unlimited
- You wait while the match is not freezed (`!InFreezeTime`)
- You trigger the unpause event when `g_PauseTimeUsed` is less than or equal to 0 (meaning the pause is finished).

scripting/get5/pausing.sp
PhlexPlexico: Hey! Thanks for taking the time review, it's much appreciated!

I think initially I wanted to keep track of this here because I had an initial assumption/testing that a timer would be created each time a person called the `!pause` command, which would create a new timer and erroneously cause another timer, and thus event firing, to be sent off when the timer finished. However, I don't think this is the case upon further review. I'll make that change and remove the handler and test that out here :) Thanks again!

PhlexPlexico:
Hey @Drarig29! 

So I think there may be an issue with pausing that I just encountered, and was wondering how to continue. I've removed the global handler from the timer counter, however I noticed that if you pause during the countdown to knifing (as an example, haven't tested without knife rounds), your pause count goes down, but the game won't pause when it moves to the next state. This will also increase the countdown timer for this functionality to *2 what it should be, and will also create additional timers, but also sends out however many `EventLogger_UnpauseCommand` calls in relation to the amount of `!pause`s called. 

I think somewhere along one of my commits, I used the global handler to check if the timer was running (and also had `g_PauseTimeUsed` under that if statement) to avoid the new event from being set off twice. However, that's more of a band-aid fix than anything.

I also tested this pause issue with the latest commit on the master branch, and it has the same issue where a team's timeout gets taken away and the game does not go into a paused. Should a separate PR/issue be created for this? If so, I won't put a possible fix in this merge.

I think this is due to the fact that when both teams are ready, and the game state flips from warmup to knife round during the warmup countdown to knife. This can be observed here

https://github.com/splewis/get5/blob/358a67a328f147770d16191888847ed8504b9a3b/scripting/get5.sp#L1314

But then `mp_warmuptime` is called for 5 seconds. 

https://github.com/splewis/get5/blob/master/scripting/get5/kniferounds.sp#L7

For these 5 seconds, a user on a team can call `!pause` and have one of their pauses taken away for the half, but not have freeze time within the knife round. Is there a possible way to check if the game is in a current countdown (would this still be `m_bFreezePeriod`?) then just return the plugin being handled again to avoid this?

Thanks!

Drarig29:
Because you tested the issue on the master branch, we are sure it's not your PR which added the bug. So yeah, IMO you should make another PR.

And sorry I'm not an expert in CS:GO timings, I'm not even perfectly familiar with the game. 

What's more, I don't have a server to do the testing anymore, I have a new PC.

I just wanted to give a review to wake up the PR 😁

Drarig29 (approved):

PhlexPlexico:
@splewis Apologies for the ping, but I was wondering if this could get reviewed/merged? It would be nice from an event-based perspective to be notified when matches get paused if organizers are running more than one match at a time.

splewis:
I'll take a look this week.

PR #650, opened 2021-03-07
Allowing customized path for round and pre-live backups.

RiadVargas:
This feature makes easier and safer exposing backups as volumes (e.g. Kubernetes deployments) and removes the necessity of an external script (rsync or something related) to move files across the filesystem.

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -83,6 +83,8 @@ ConVar g_TimeFormatCvar;
 ConVar g_VetoConfirmationTimeCvar;
 ConVar g_VetoCountdownCvar;
 ConVar g_WarmupCfgCvar;
+ConVar g_RoundBackupFormatCvar;
+ConVar g_PreLiveBackupFormatCvar;
 
 // Autoset convars (not meant for users to set)
 ConVar g_GameStateCvar;
@@ -345,6 +347,12 @@ public void OnPluginStart() {
                    "Seconds to countdown before veto process commences. Set to \"0\" to disable.");
   g_WarmupCfgCvar =
       CreateConVar("get5_warmup_cfg", "get5/warmup.cfg", "Config file to exec in warmup periods");
+  g_RoundBackupFormatCvar = CreateConVar(
+      "get5_backup_format", "get5_backup_match%s_map%d_round%d.cfg",
+      "Path where round backups will be stored.");
+  g_PreLiveBackupFormatCvar = CreateConVar(
+      "get5_backup_format", "get5_backup_match%s_map%d_prelive.cfg",
+      "Path where pre-live backups will be stored.");
 
   /** Create and exec plugin's configuration file **/
   AutoExecConfig(true, "get5");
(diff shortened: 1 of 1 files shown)


nickdnk:
I don't know how paths are resolved, to be honest. Have you tested that this works with absolute paths outside the csgo directory?

RiadVargas:
> I don't know how paths are resolved, to be honest. Have you tested that this works with absolute paths outside the csgo directory?

@nickdnk Since the underlying function when saving backups is `smn_KeyValuesToFile` (ref. https://github.com/alliedmodders/sourcemod/blob/master/core/smn_keyvalues.cpp#L754), it supports relative paths within `csgo` directory.

I haven't tried to escape base filesystem, by using ".." for instance, though.

scripting/get5.sp
```diff
@@ -345,6 +347,12 @@ public void OnPluginStart() {
                    "Seconds to countdown before veto process commences. Set to \"0\" to disable.");
   g_WarmupCfgCvar =
       CreateConVar("get5_warmup_cfg", "get5/warmup.cfg", "Config file to exec in warmup periods");
+  g_RoundBackupFormatCvar = CreateConVar(
+      "get5_backup_format", "get5_backup_match%s_map%d_round%d.cfg",
+      "Path where round backups will be stored.");
+  g_PreLiveBackupFormatCvar = CreateConVar(
+      "get5_backup_format", "get5_backup_match%s_map%d_prelive.cfg",
```
nickdnk, line 354: I don't think we can have two convars with the same name. Both are `get5_backup_format` here.

scripting/get5.sp
```diff
@@ -1155,13 +1163,14 @@ public void WriteBackup() {
     return;
   }
 
+  char backupFormat[PLATFORM_MAX_PATH];
   char path[PLATFORM_MAX_PATH];
   if (g_GameState == Get5State_Live) {
-    Format(path, sizeof(path), "get5_backup_match%s_map%d_round%d.cfg", g_MatchID,
-           GetMapStatsNumber(), GameRules_GetProp("m_totalRoundsPlayed"));
+    g_RoundBackupFormatCvar.GetString(backupFormat, sizeof(backupFormat));
+    Format(path, sizeof(path), backupFormat, g_MatchID, GetMapStatsNumber(), GameRules_GetProp("m_totalRoundsPlayed"));
```
nickdnk, line 1170: Note to myself to use `g_RoundNumber` here after 0.9 is merged.

nickdnk:
This is alright, but please rebase it on master and then add these options to the documentation.

Here: get5_backup_format

nickdnk:
I just tried implementing this, and the problem is that `get5_listbackups` uses part of the format to list backups. I'm not sure how we can implement this. `get5_listbackups` might need to take another parameter that defaults to a prefix we could define via cvar.

nickdnk:
Closing as addressed in https://github.com/splewis/get5/pull/793.

PR #652, opened 2021-03-23
Match restore issues

arildboifot:
While testing the restore functionality I ran into 3 separate issues:

1. Restoring to another map than the current map results in the match starting Warmup or Knife depending on the current map number.
2. Restoring to second half on an another map than the current map results in the "end of first half screen" and incorrect match scores. Also the teams may be spawned on incorrect sides.
3. mp_match_restart_delay was not set

Any feedback is much appreciated!

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -670,6 +670,12 @@ public Action Timer_CheckReady(Handle timer) {
   if (g_GameState == Get5State_None) {
     return Plugin_Continue;
   }
+  
+  if (g_DoingBackupRestoreNow) {
+    LogDebug("Timer_CheckReady: Waiting for restore");
+    return Plugin_Continue;
+  }
+
 
   CheckTeamNameStatus(MatchTeam_Team1);
   CheckTeamNameStatus(MatchTeam_Team2);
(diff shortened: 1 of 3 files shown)


arildboifot:
There's still a issue that results in an unpaused match when restoring on a clean server for the first time. When restoring a second time everything is good. I'll try to dig into that tomorrow.

nickdnk:
@arildboifot Do you have any updates here? Seems like you are onto something here. Come on the Discord if you want. Link in readme.

rpkaul:
@arildboifot  you have any update on this brother ?

arildboifot:
Thanks for reminder! We've been running with this modification since this PR and done a quite a few backs/restores without hitting the edge case described above. I haven't had time to look more into the edge case unfortunately.

nickdnk:
@splewis Since @arildboifot says it works, could you take a look and merge it if you don't see any issues? I'm not very familiar with backups in Get5, as I don't use them at all, so I'd rather not merge this alone.

splewis:
I'll take a look at this during this weekend.

PR #686, opened 2021-07-17
Stats in JSON format

jenrik:
Add support for writing stats in JSON format.

The feature is enabled by setting the `get5_stats_path_format` cvar to a value ending in `.json`

_The work is paid for in part by [ehub.gg](https://ehub.gg/)_

### DIFF
--- a/scripting/get5/stats.sp
+++ b/scripting/get5/stats.sp
@@ -514,23 +514,39 @@ static int GetClutchingClient(int csTeam) {
   }
 }
 
-public bool DumpToFile() {
+public void DumpToFile() {
   char path[PLATFORM_MAX_PATH + 1];
   if (FormatCvarString(g_StatsPathFormatCvar, path, sizeof(path))) {
-    //g_StatsKv.ExportToFile(path);
-    
-    return DumpToFilePath(path);
-  } else {
-    return false;
+    DumpToFilePath(path);
   }
 }
 
 public bool DumpToFilePath(const char[] path) {
+  int path_length = strlen(path);
+  // Check if path ends in ".json"
+  if (path_length >= 5) {
+    char ext[5];
+    ext[0] = path[path_length-5];
+    ext[1] = path[path_length-4];
+    ext[2] = path[path_length-3];
+    ext[3] = path[path_length-2];
+    ext[4] = path[path_length-1];
+    if (StrEqual(".json", ext, false)) {
+      // Dump in json format
+      return DumpToJSONFile(path);
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


scripting/get5/stats.sp
```diff
@@ -517,8 +517,84 @@ static int GetClutchingClient(int csTeam) {
 public void DumpToFile() {
   char path[PLATFORM_MAX_PATH + 1];
   if (FormatCvarString(g_StatsPathFormatCvar, path, sizeof(path))) {
-    g_StatsKv.ExportToFile(path);
+    DumpToFilePath(path);
+  }
+}
+
+public bool DumpToFilePath(const char[] path) {
```
splewis, line 524: A suggestion:

* Add a generic `IsJSONPath` function  to util.sp
* That implementation can be simplified doing something like: `StrCompare(path[path_length - 5], ".json") == 0` (after validating path_length > 5 still)
* Replace [this](https://github.com/splewis/get5/blob/fae6a4f76eb042cb555cc8b33ed67ebb0d87cb45/scripting/get5/matchconfig.sp#L149) with a call to that function
* Make this function: `return IsJSONPath(path) : DumpToJSONFile(path) :  g_StatsKv.ExportToFile(path);`

scripting/get5/stats.sp
```diff
@@ -517,8 +517,84 @@ static int GetClutchingClient(int csTeam) {
...
+  JSON_Object stats = EncodeKeyValue(g_StatsKv);
+  g_StatsKv.Rewind();
+
+  File stats_file = OpenFile(path, "w");
+  if (stats_file == null) {
+    LogError("Failed to open stats file");
+    return false;
   }
+
+  char jsonBuffer[8192]; // 8 KiB
```
splewis, line 556: I'm not sure how large the json data is in practice TBH, I'm wondering if 8KiB is enough; do you have a sense of the size for a normal 10-person 3-map match?

jenrik: This was a guess based on intuition. I tried 32 KiB but that crashed with a out of heap or stack space error (I don't recall which)

scripting/get5/stats.sp
jenrik: Implemented in a3e9d2b6323fa5df4fae241eb0753aadf3d4f820

ErcouldnT:
Hi, sir

I have just started [this project](https://github.com/ErcouldnT/Get5-Match-Stats-API) about JSON match stats.

If you are able to send match stats JSON body as a POST request to a server set with get5_match_stats_api_url that would be awesome!

Erkut

Issue #504, opened 2020-04-20
Freeze Players during Veto


TandelK:
Lot of time people are just moving around in Warmup. If we could implement Freeze Players function during Veto selections. With this random people getting Veto would focus on Veto rather than by mistake picking up map for Weapon Selection.

Drarig29:
Maybe you already know how to do it, but I found a way [here](https://forums.alliedmods.net/showthread.php?p=2438705) with his command `sm_speed`:

From his source file:

```
SetEntPropFloat(entity, Prop_Data, "m_flLaggedMovementValue", value);
```

splewis:
I'd accept a PR for this.

> SetEntPropFloat(entity, Prop_Data, "m_flLaggedMovementValue", value);

I think using the builtin pausing or freezetime functionality of the game is a vastly better way to implement this. Awkwardly freezing players would probably not be something I'd accept.

TandelK:
@Drarig29 Try to integrate this feature that as soon as Veto is called upon all the players gets freeze.

Drarig29:
I don't have time currently. Maybe in the next weeks.

TandelK:
As this is integrated now , Closing this issue.

PR #696, opened 2021-08-25
Pause Game While Veto is going on. 

TandelK:
Just a simple cvars to added to pause game during Veto time. This allow players to focus properly on Veto

Add for https://github.com/splewis/get5/issues/504


scripting/get5/mapveto.sp
```diff
@@ -14,6 +14,15 @@ public void CreateVeto() {
   g_VetoCaptains[MatchTeam_Team1] = GetTeamCaptain(MatchTeam_Team1);
   g_VetoCaptains[MatchTeam_Team2] = GetTeamCaptain(MatchTeam_Team2);
   ResetReadyStatus();
+  if (g_FreezePlayersVetoCvar.BoolValue)
+  {
+    Pause();
+    for (int i = 1; i <= MaxClients; i++) {
```
Drarig29, line 20: For the sake of reusing what's existing, you could use `LOOP_CLIENTS(i) {` instead of `for (int i = 1; i <= MaxClients; i++) {`.

scripting/get5/mapveto.sp
```diff
@@ -42,6 +51,15 @@ public void VetoFinished() {
   ChangeState(Get5State_Warmup);
   Get5_MessageToAll("%t", "MapDecidedInfoMessage");
 
+  if (g_FreezePlayersVetoCvar.BoolValue) {
+    Unpause();
+    for (int i = 1; i <= MaxClients; i++) {
```
Drarig29, line 56: Same here 😊

Drarig29 (commented):
Glad you did it, I didn't find the time to reinstall a server and do it on my own ^^

TandelK:
@Drarig29 I forgot the main part of the whole update to declare the ConVar first. 🤣 . Removed Entity color also and updated it to use LOOP_CLIENTS now as per review. 

@splewis Check accordingly. The reason why I used Pause is because the players who gets disconnected in  between of the Veto can again spawn and move and same can be overcome by Respawning as they are frozen. So Pause the match and than freeze them and once Veto is over the match will again become unpause state and unfreeze all the players.

TandelK:
@splewis Can check it once ?

splewis:
I'm not a huge fan of freezing players this way, as it's fairly cumbersome from a player POV, so I'm hesitant to merge this before I actually  test it out ingame myself.

I'd be more open if there was a way to freeze players more naturally; e.g. a round restart and pause the game. That may require some changes to end the warmup phase, if we're in it though.

TandelK:
@splewis I can remove the Freeze Player Entity and by default it works in the way you told. As soon as Warmup ends it restarts the round and than pauses the match and once Veto is over Unpause command is executed directly. 

https://github.com/splewis/get5/blob/44171af576cff9c56619d936564d562d747642f8/scripting/get5/mapveto.sp#L17-L25

TandelK:
@splewis Have updated as per last message , i have removed freezing player entity but instead i have now used simple pause and unpause events with round restart as soon as Warmup is over.

nickdnk:
@TandelK Can you rename the cvar accordingly, since it's no longer a freeze?

scripting/get5/mapveto.sp
```diff
@@ -14,6 +14,8 @@ public void CreateVeto() {
   g_VetoCaptains[MatchTeam_Team1] = GetTeamCaptain(MatchTeam_Team1);
   g_VetoCaptains[MatchTeam_Team2] = GetTeamCaptain(MatchTeam_Team2);
   ResetReadyStatus();
+  if (g_FreezePlayersVetoCvar.BoolValue) {
+    Pause(); }
```
nickdnk, line 18: @TandelK And fix format here. Bracket down a line. Same in VetoFinished below.

TandelK: > @TandelK Can you rename the cvar accordingly, since it's no longer a freeze?

Its still technical Freezing right or should i rename it to
`g_PausePlayersVetoCvar.BoolValue`

TandelK: @nickdnk I have resolved both the above mentioned comments.

nickdnk:
@TandelK Wouldn't it theoretically cause an infinite pause if you changed the Cvar after veto begins? Couldn't we call Unpause regardless of the Cvar status, as unpausing when there is no pause does nothing? Just trying to think how to avoid bugs here.

nickdnk:
Also please add the CVar here: https://github.com/nickdnk/get5/blob/master/scripting/get5/debug.sp#L140 - we need to keep that list updated with new stuff we put in.

TandelK:
> @TandelK Wouldn't it theoretically cause an infinite pause if you changed the Cvar after veto begins? Couldn't we call Unpause regardless of the Cvar status, as unpausing when there is no pause does nothing? Just trying to think how to avoid bugs here.

This is how it works 
As soon as Warmup ends for VETO , map round restarts and pause is triggered. This is cvar not a cmd that has to be executed while VETO is on going. As soon as VETO End is triggered , the unpause event is triggered. Also CVAR in debug is not required as it is not compulsory CVAR.

nickdnk:
> > @TandelK Wouldn't it theoretically cause an infinite pause if you changed the Cvar after veto begins? Couldn't we call Unpause regardless of the Cvar status, as unpausing when there is no pause does nothing? Just trying to think how to avoid bugs here.
> 
> 
> 
> This is how it works 
> 
> As soon as Warmup ends for VETO , map round restarts and pause is triggered. This is cvar not a cmd that has to be executed while VETO is on going. As soon as VETO End is triggered , the unpause event is triggered. Also CVAR in debug is not required as it is not compulsory CVAR. I can still work on adding a command to check if game is paused while veto is finished and unpause it to add that exception issue that you just mentioned
> 
> 
> There is one known issue which i found while triggering this , If someone has removed Pause functionality for using 3rd party plugin , it might not work for them. I would have to rewrite Pause() and Unpause() with Server Command instead of using functions created by Get5.

My point is that just because the Cvar is not changed by get5 in this context does not mean it can't be. Someone could mess with pauses while the veto is going on, so I just suggested that you always call Unpause, as it is without consequences if the game isn't paused.

What do you mean it's not "required" in the debug? Why not put it there? I'm not saying you *have to* for anything to work, just that we might as well make the debug as if informative as possible if we're adding variables that affect game state.

scripting/get5.sp
```diff
@@ -310,6 +311,9 @@ public void OnPluginStart() {
   g_ResetPausesEachHalfCvar =
       CreateConVar("get5_reset_pauses_each_half", "1",
                    "Whether pause limits will be reset each halftime period");
+  g_PauseOnVetoCvar =
+      CreateConVar("get5_pause_on_veto", "0",
+                   "Set 1 to Pause Match during Veto time}");
```
nickdnk, line 316: Please remove the lingering `}`.

TandelK: Updated

TandelK:
> I'm not a huge fan of freezing players this way, as it's fairly cumbersome from a player POV, so I'm hesitant to merge this before I actually test it out ingame myself.
> 
> I'd be more open if there was a way to freeze players more naturally; e.g. a round restart and pause the game. That may require some changes to end the warmup phase, if we're in it though.

@splewis  Have updated it as per this converstation, now its pausing the game after warmup ends and after Veto is called by server and called Paused. If Pausing CVAR is also disable this can use in game Pause System to achieve this effect. Also have added after Veto complete its going to use IsPaused() query just in case someone uses manual game pause command also after veto is called and unpause the game.

TandelK:
@splewis Just have a look at it once and let me do know if we can merge it.

Issue #685, opened 2021-07-15
Failed to get matchid from match init query


nikitosoz8:
### Expected behavior

Match ID gets from the json file. Data saved in db.

### Actual behavior

L 07/15/2021 - 10:24:45: [get5.smx] Failed to read input auth id "", inputId
L 07/15/2021 - 10:24:45: [get5_mysqlstats.smx] Failed to get matchid from match init query

### Steps to reproduce

Here is match config:
`{
"matchid": "6373",
"num_maps": 5,
"players_per_team": 1,
"min_players_to_ready": 1,
"min_spectators_to_ready": 0,
"side_type": "never_knife",
"skip_veto": true,
"maplist":["aim_9h_ak","aim_ak47_v2","aim_map2_go","aim_map_cl","aim_map"],
"spectators":{"players":[""]},
"team1":{
"name":"mjk134","players":{"STEAM_0:0:531076147":"mjk134"}},
"team2":{"name":"nikitosoz8","players":{"STEAM_0:1:175436101":"nikitosoz8"}}
}`

Insert the config in a json file, load it on the server and start the match. 

- Plugin version: 0.7.3-dev+529
- Sourcemod version: 1.10.0.6502

nickdnk:
Will be fixed with #673.

Issue #453, opened 2020-01-15
[BUG] "suicide" value on get5_apistats is not correct


FlowingSPDG:
Hi, Im developing my own get5-web by Golang ([get5-web-go](https://github.com/FlowingSPDG/get5-web-go)).
and I found bug while im developing it.
When player killed themself by using HEgrenade/Molotov,it will be counted as "suicide" because killer and victim is equal.
but when player killed by "World"(such as falling,especially vertigo) get5 won't recognize "World" as valid killer so it won't counted as "suicide".
https://github.com/splewis/get5/blob/95e52f5215d6aa6d576a89a1ca3c9064edba0a03/scripting/get5/stats.sp#L198

Thanks for best match plugin btw :)
https://github.com/splewis/get5/blob/95e52f5215d6aa6d576a89a1ca3c9064edba0a03/scripting/get5/stats.sp#L228

https://github.com/FlowingSPDG/get5-web-go/issues/67

vassilidev:
Thx !!

FlowingSPDG:
❤️‍🔥

PR #729, opened 2022-01-02
version 0.9

nickdnk:
1. Added new forwards for player killed, round end, round started (freeze time ends), grenade thrown and MVP election.
2. Added bomb plant and defuse forwards, including time remaining (milliseconds) for defuse.
3. Cleaned up some suicide logic, should fix https://github.com/splewis/get5/issues/453
4. Add a few missing LogDebug calls.
5. Changed some properties and hooks to alphabetical order.
6. Removed unused `player_team` event hook.
7. Made MatchID a string everywhere (no redeclaration in apistats or mysqlstats). It now defaults to an empty string and not "matchid".
8. Removed the `get5_mysql_force_matchid` Cvar and fixed the code that sets it from the match config. Should resolve https://github.com/splewis/get5/issues/685

**Breaking changes:**

I plan on this being a breaking change release, so I made these adjustments. I track all the breaking changes here and I suggest we add a changelog.

1. Changed winReason to match the enumerations (CSGO added 1 to these values for some reason). See https://github.com/saul/demofile/issues/38#issuecomment-330068048
2. Changed event data structure for `player_death`, adding the entire `assist` as an object that can be null if there was no assist. It's cleaner this way, JSON-wise.
3. Player death event now fires for friendly fire and suicides and includes assist-friendly fire as well. If you don't want this data you have to now check for these booleans.
4. Removed `get5_web_avaliable` legacy spelling-mistake command alias.
5. stock `HelpfulAttack` no longer checks for `isValidClient` - you should do this before using the method. This because it caused duplicate calls to `isValidClient` in most use-cases.
6. `matchId` is now passed as the first `char[]` argument to most forwards where it makes sense to have it, instead of those forwards having to call `GetMatchID` immediately after. The order is now always `matchId`, `mapNumber`, `roundNumber` and `roundTime` for all forward events where these are applicable. Please check your argument order in forwards!
7. Added `round_time` and `round_number` to applicable events and forwards. `round_time` is milliseconds and `round_number` is zero-indexed like map number.

If anyone has any feedback please comment. I did not test this code yet so there may be errors.

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -24,6 +24,7 @@
 #include <cstrike>
 #include <json>  // github.com/clugg/sm-json
 #include <sdktools>
+#include <sdkhooks>
 #include <sourcemod>
 #include <testing>
 
@@ -138,6 +139,12 @@ bool g_SavedValveBackup = false;
 bool g_DoingBackupRestoreNow = false;
 
 // Stats values
+StringMap g_FlashbangContainer; // Stores info about which flashbang entities flashed which players.
+StringMap g_HEAndDecoyGrenadeContainer; // Stores info about which HE grenades and decoys hit which players, as they share damage type.
+StringMap g_MolotovContainer; // Stores info about which players were damaged by which molotov grenades.
+StringMap g_SmokeGrenadeContainer; // Stores info about which players threw which smokes and if the extinguished molotovs.
+int g_LatestUserIdToDetonateMolotov = 0; // Molotov detonate and start-burning/extinguish are two separate events always fired right after each other. We need this to bind them together.
+int g_LatestSmokeGrenadeToDetonateOnMolotov = 0; // Used to attribute extinguish booleans to smoke grenades.
 bool g_SetTeamClutching[4];
 int g_RoundKills[MAXPLAYERS + 1];  // kills per round each client has gotten
 int g_RoundClutchingEnemyCount[MAXPLAYERS +
@@ -207,6 +214,11 @@ Handle g_OnBombDefused = INVALID_HANDLE;
 Handle g_OnBombPlanted = INVALID_HANDLE;
 Handle g_OnDemoFinished = INVALID_HANDLE;
 Handle g_OnEvent = INVALID_HANDLE;
+Handle g_OnFlashbangDetonated = INVALID_HANDLE;
+Handle g_OnHEGrenadeDetonated = INVALID_HANDLE;
+Handle g_OnSmokeGrenadeDetonated = INVALID_HANDLE;
+Handle g_OnDecoyEnded = INVALID_HANDLE;
+Handle g_OnMolotovEnded = INVALID_HANDLE;
 Handle g_OnGameStateChanged = INVALID_HANDLE;
 Handle g_OnGoingLive = INVALID_HANDLE;
 Handle g_OnGrenadeThrown = INVALID_HANDLE;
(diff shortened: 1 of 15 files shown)


scripting/get5/stats.sp
```diff
@@ -296,6 +329,54 @@ public Action Stats_PlayerDeathEvent(Event event, const char[] name, bool dontBr
...
+    weapon,
+    friendlyFire,
+    assistFriendlyFire,
+    penetrated,
+    thruSmoke,
+    noScope,
+    attackerBlind
+  );
+
+  LogDebug("Calling Get5_OnPlayerDied(weapon=%s, headshot=%d, mapNumber=%d, roundNumber=%d, roundTime=%d, attacker=%d, victim=%d, assister=%d, assistedFlash=%d, penetrated=%d, thruSmoke=%d, noScope=%d, attackerBlind=%d, attackerTeam=%d, assisterTeam=%d, victimTeam=%d)",
```
PhlexPlexico, line 358: Maybe make the event `Get5_OnPlayerDeath`?

PhlexPlexico:
Aside from the one event name this looks good to me. I would personally like to test this as well with the integrations (G5API, etc) I have as well. I know there are breaking changes but from the looks of it, I think there shouldn't be any big reasons why web panels shouldn't still be able to work with these new features. I could probably do some testing in the coming weeks here as well if needed (nothing too extensive, but I think I could get together some 1v1/2v2 testing).

scripting/get5.sp
```diff
@@ -112,6 +112,10 @@ char g_FavoredTeamText[MAX_CVAR_LENGTH];
 int g_PlayersPerTeam = 5;
 int g_MinPlayersToReady = 1;
 int g_MinSpectatorsToReady = 0;
+int g_RoundNumber = 0;
```
splewis, line 115: This value is likely to be incorrect with backup/restores. I would avoid storing it at all, and derive it otherwise (there are some netprops to calculate it, used somewhere else in the plugin).

scripting/get5/eventlogger.sp
```diff
@@ -37,6 +37,8 @@ static void EventLogger_LogEvent(const char[] eventName, JSON_Object params) {
     Call_PushString(buffer);
     Call_Finish();
   }
+
+  json_cleanup_and_delete(json);
```
splewis, line 41: That was missing, oh my :)

scripting/get5/stats.sp
splewis: +1

holdzin:
POALWBDHJDNAJSJSNSNSNJAEK

Em dom., 2 de jan. de 2022 às 19:29, Sean Lewis ***@***.***>
escreveu:

scripting/include/get5.inc
```diff
@@ -81,6 +81,26 @@ forward void Get5_OnMapVetoed(MatchTeam team, const char[] map);
...
 
+// Called when a player is elected the MVP of the round.
+forward void Get5_OnPlayerBecameMVP(int mapNumber, int roundNumber, int client, int clientSide, int reason);
+
+// Called when a player is killed in a live match. Used by MySQL stats. roundTime is milliseconds.
+forward void Get5_OnPlayerDied(char[] weapon, bool headshot, int mapNumber, int roundNumber, int roundTime,
+    int attacker, int victim, int assister, bool flashAssist, int penetratedObjects, bool thruSmoke, bool noScope,
+    bool attackerBlind, int attackerSide, int assisterSide, int victimSide);
+
+// Called when a round starts. Used by MySQL stats.
```
splewis, line 92: No need for comments like "Used by MySQL stats", imo. None of the public API being exposed here is meant to be specific to any particular use.

nickdnk:
@PhlexPlexico Note I just added a suicide boolean to player death. This because suicide can happen both if attacker == victim *or* if attacker is invalid (world damage). So to simplify I calculate this boolean and include it in both the forward and event.

PhlexPlexico:
Perfect. I think on the web side I don't actually even touch the `OnPlayerDeath` event, but this is getting seemingly close to allowing for live updates on a web panel now which is awesome to see!

nickdnk:
I think it makes more sense to pass the matchId as an argument to all the forwards instead of using the GetMatchID method. I'm going to refactor that.

PhlexPlexico:
So there's one more issue actually that I'd like to have tested since I've had to do a hacky workaround for the API Key and Web URL values, and I'm unsure if it will have to be done with the match ID... Currently, when a demo finishes and the forward `Get5_OnDemoFinished` is called, there is actually zero information on the previous API key and URL that was given to get5 since it already deletes those values when the match finishes. Will this be the same case now for the match ID? For example, when I wish to upload a demo I actually have to store the API key and URL in different variables within my plugin in order to use them properly:

https://github.com/PhlexPlexico/G5WS/blob/986955db4e2d8ee4abec2505117154019cc9dfc6/scripting/G5WS.sp#L439-L443

Will this be the same case now with the match ID?

nickdnk:
I'm unsure of this. I have never used that forward as I don't record demos. I just added the matchId as argument. What value is in it will be unchanged. You're going to have to test and get back to me.

nickdnk:
I'm closing this as I'm working on some major changes related to the other PR I opened.

PR #733, opened 2022-01-06
Add forwards for flashbang and grenade victims

nickdnk:
Hello

This is a feature planned for after the 0.9 branch has been merged. It provides information about who were flashed or hit by a grenade (by whom) and for how long they were flashed or how much they were damaged, respectively. I have not yet tested this code, but it compiles and *should* work.

~~Due to how molotov grenades work, it's not really possible to track their damage this way.~~ Edit: I figured out how to do this. I added the skeleton for it and will implement it soon.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.zip
 *.tar.gz
 builds
+.idea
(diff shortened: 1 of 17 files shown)


nickdnk:
Closed here as the base has to be master and not 0.9

Issue #707, opened 2021-10-17
Add an option to not cap damage at 100


GekasD:
Why is this even a thing in the first place? it feels like the current behavior should be optional and turned off by default, or at least have a convar to disable it. I can do a PR to implement this if  you think its worth being added.

nickdnk:
Why would you need this information? It's meaningless. I assume you would only want this to apply to the damage report at each round end, because the actual damage dealt is a metric for your impact and is always capped at 100, even in the official game. You can see this in the new premier matchmaking where the column called Damage is on the scoreboard. It does not give you 450 damage if you headshot someone with an AWP.

rpkaul:
@nickdnk I agree its not that important to see total given damage but if its possible to see more information then i love to see that, its fun to see how much damage we did last round , what was the total damage on 1 player. thats the main reason i dont want to cap this .

nickdnk:
Excess damage doesn't count towards ADR or damage given. I'd argue it makes things more confusing and inconsistent when taking the other stats we collect into consideraton.

nickdnk:
Add in https://github.com/splewis/get5/pull/736.

stalsus:
I'm glad this is added, but I still believe it should also be included in the stat tracking, kills * 100 is useless in my opinion.

nickdnk:
> I'm glad this is added, but I still believe it should also be included in the stat tracking.

Nowhere else is damaged tracked like this. It gives a completely skewed impression of damage given across multiple rounds.

stalsus:
Fair enough, thanks for adding this 👍🏻

Issue #695, opened 2021-08-20
[Feature Request] Add all Individual Player Weapon Information , API to Web Panel or MySQL Stats plugin. 


TandelK:
We should store all the individual player weapon kills information for brief information . We can create core inside the Get5 Plugin but can access for storing via API Panel and MySQL Stats plugin. 

Every individual weapon kills can be stored for brief data and can be used via API Panel or MySQL Plugin which can show data using 3rd party web app.

nickdnk:
I'm adding some forwards soon that will allow people to get this information.

nickdnk:
This should be possible with https://github.com/splewis/get5/pull/736. Please try it out.

Issue #673, opened 2021-06-07
get5_server_id isn't saved according to the convar


astutejoe:
### Expected behavior
Entries in get5_stats_matches table (MySQL plugin) should be inserted with server_id being the get5_server_id ConVar, that ConVar can be set on get5.cfg (csgo/cfg/sourcemod/get5.cfg). I also tried setting it manually through the console by issuing get5_stats_matches 5 and also tried through the cvars section of the match.cfg

### Actual behavior
server_id is always saved as 0

### Steps to reproduce

- Plugin version: 0.7.1
- Sourcemod version: 1.10.0.6502
- Steps to reproduce (please be specific):
Change get5_server_id to something other than 0
Start the server
Watch an entry be created in the MySQL server with server_id = 0.

Issue #440, opened 2019-11-11
Mapnumber in mysql_stats not increased after tie


allanwoer:
If making a feature request, you should delete all the pre-filled text here.
If reporting a bug, fill in the following sections:

### Expected behavior
mysql_stats:
After match tie, mapnumber should be increased, so match stats will be updated correctly.

### Actual behavior
Mapnumer is not increased after a match tie. Data from the previous match is overwritten after a match tie. Seems like mapnumber is calculated by team score, and not matches played.

### Steps to reproduce
- Plugin version: 0.7.1
- Sourcemod version: 1.9.0.6276
- Steps to reproduce (please be specific): End a match with more than one match in the series with "TIE" as result. Note that mapnumber isn't increased.

mysql_stats event: 11/10/2019 - 21:38:32: [get5_mysql] UPDATE `get5_stats_maps` SET team1_score = 1, team2_score = 0 WHERE matchid = 88 and mapnumber = 0

get5_event: 11/10/2019 - 21:38:32: [get5] get5_event: {"matchid`":"88","params":{"reason":1,"winner":"team1","team2_score":0,"map_number":1,"team1_score":1,"map_name":"de_dust2","winner_side":"T"},"event":"round_end"}

note the difference in mapnumber.

nickdnk:
Should be fixed on master.

PR #736, opened 2022-01-11
[Major] Version 0.9

nickdnk:
Hello

After adding to my recent two PRs (that I have now closed), I have come to the conclusion that they were too tightly coupled.

So, here goes a new branch. I have squashed **a lot** of commits as well.

1. The event and forward systems have been completely rewritten to run off objects based on `methodmap`s. These are actually `StringMap`s wrapped in `JSON_Object`s and `JSON_Array`s as defined in `sm-json` and behave similarly to classes/objects as we know from modern OOP. They have constructors and offer compile-time checking (and autocomplete if your editor can figure that out for SourcePawn), meaning you are a lot more likely to catch errors without even having to run the plugin. This means:
    -  There is now always only one argument for a forward, and that's a `methodmap` subclass of `Get5Event`. No more counting `Param_Cell` and/or `Param_String` arguments, checking where you've messed up.
    -  The `EventLogger` system now directly receives these objects and logs them out, meaning there is greater consistency of key/value names, as the methodmaps inherit from each other. This is not perfect, but it's better. It also means a lot less code in the `EventLogger` "class".
    -  We can add data to events (and inherently, to forwards) without introducing breaking changes in the forward declarations.
2. As mentioned in https://github.com/splewis/get5/pull/729, I added a lot of new forwards with corresponding new event objects, such as:
    - `Get5PlayerDeathEvent` - player killed, who/what killed them, assist, flash assist and weapon info. Closes https://github.com/splewis/get5/issues/695.
    - `Get5HEDetonatedEvent` - who threw which HE and whom did they damage for how much (and if they killed their target(s)).
    - `Get5DecoyStartedEvent` - who started which decoys.
    - `Get5GrenadeThrown` - all grenades thrown (but no victim or duration info). Triggered when the grenade is thrown, not when it detonates, and it can distinguish between molotov and incendiary. It does not contain information about victims.
    - `Get5FlashbangDetonatedEvent` - who threw which flashes and whom did they blind for how long.
    - `Get5MolotovDetonatedEvent` - who threw which mollies and who were damaged by them and for how long did they burn, and if they died. Sends all events as "molotov" as it *cannot* distinguish between molotov and incendiary. **NOTE:** This event does *not* fire if the molotov is thrown directly at a smoke (and hence does nothing). Also, it fires when the molotov ends, not when it detonates, but its `RoundTime` parameter reflects time of detonation.
    - `Get5SmokeDetonatedEvent` - who threw which smokes and did they extinguish any molotovs.
    - `Get5BombDefusedEvent`, `Get5BombPlantedEvent` and `Get5BombExplodedEvent` - who planted/defused the bomb, at which site and how long was left when defusing.
    
        **get5.inc describes these objects and is sort of self-documenting (if such a thing exists).**

3. I fixed some bugs:
    - Fix https://github.com/splewis/get5/issues/685 (also removed the `get5_mysql_force_matchid` Cvar as it is defined in the match config)
    - Fix https://github.com/splewis/get5/issues/453
    - Fix https://github.com/splewis/get5/issues/673 (I could not replicate this bug, but I have verified that this function works on this branch)
    - Fix https://github.com/splewis/get5/issues/2 and fix https://github.com/splewis/get5/issues/1 by simply letting the team menu remain as is, but putting the player on the correct team when joining _if_ there are no pending team swaps. Essentially it's a middle-of-the-road kind of solution that *should* give us what we want without breaking things.
    - Fixed memory leak in the eventlogger: https://github.com/splewis/get5/blob/8ffecad946c8206494fdd3ca856d37d1394f1757/scripting/get5/eventlogger.sp#L1 - the `new JSON_Object()` was never deleted.
    - Fix https://github.com/splewis/get5/issues/440
4. I adjusted `get5_mysql` and `get5_apistats` examples so they should work with these changes. Still needs testing.
5. All `g_xxxxCvar` handles are now correctly suffixed with `Cvar`.
6. `g_MatchID` is now always handled as a `char[]` and no longer duplicated as an `int` in `get5_mysql`.
7. Removed duplicate `GetMapNumber()` logic defined in both `get5_mysql` and `get5_apistats`. There is now a native called `Get5_GetMapNumber()` that does the same thing.
8. We now use the `SDKHook_OnTakeDamageAlive` hook instead of the `player_hurt` event, as it provides the entity that damaged the player. We need that to resolve grenade damage for the new events.
9. `CSRoundEndReason` has been decremented by 1 in the `Get5RoundEndedEvent` to match the values defined in the enumeration. For some reason, CSGO has these incremented by 1: https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/cstrike.inc#L53-L77
10. Added `get5_print_damage_excess` option which lets you configure the damage report to print the full, uncapped damage in chat on round end. This does *not* affect any other metrics than chat and is disabled by default. Fix https://github.com/splewis/get5/issues/707
11. In order to reliably fire events that happen after a map is decided, such as a grenade going off, someone falling to their death or someone typing in chat, I had to introduce a `g_MapNumber` variable that works similarly to `g_RoundNumber` and is fired on round start, meaning after-time occurrences will have the correct map number always.

4. New documentation of the events and forwards using the OpenAPI Swagger spec.

5. Get5 now requires SourceMod 1.10 as [sm-json](https://github.com/clugg/sm-json) was updated to v4. Also removed `json_load_file` as sm-json now provides this function (https://clugg.github.io/sm-json/#content-global-methods-json-read-from-file). Fixed a bunch of memory management issues while cleaning this up.

This is still a work in progress.


nickdnk:
Also, https://github.com/clugg/sm-json/pull/23 needs to be merged to avoid knowingly introducing memory leaks on grenade events, as we snapshot these objects. @clugg

deLuXe83:
Hey :)
What do you think - when will this version be merged into master?

Thx!!

nickdnk:
> Hey :)
> 
> What do you think - when will this version be merged into master?
> 
> 
> 
> Thx!!

Not sure. I'm waiting for some feedback on the PR. I won't merge it until someone looks it over.

With that said it does seem to work for what I use it for.

rpkaul:
i  tested this version and found few bugs
1 - auto team select on connect is not working on map change after veto
2- after map change i can join any team , steam auth condition is not wotking 
3- !tech command working as !pause ( timeout)

scripting/get5.sp
```diff
@@ -1168,11 +1253,11 @@ public void WriteBackup() {
   char path[PLATFORM_MAX_PATH];
   if (g_GameState == Get5State_Live) {
     Format(path, sizeof(path), "get5_backup_match%s_map%d_round%d.cfg", g_MatchID,
-           GetMapStatsNumber(), GameRules_GetProp("m_totalRoundsPlayed"));
+           GetMapStatsNumber(), GetRoundsPlayed());
   } else {
-    Format(path, sizeof(path), "get5_backup_match%s_map%d_prelive.cfg", g_MatchID,
-           GetMapStatsNumber());
+    Format(path, sizeof(path), "get5_backup_match%s_map%s_prelive.cfg", g_MatchID);
```
PhlexPlexico, line 1258: Missing `GetMapStatsNumber() in call`

nickdnk:
> i tested this version and found few bugs 1 - auto team select on connect is not working on map change after veto 2- after map change i can join any team , steam auth condition is not wotking 3- !tech command working as !pause ( timeout)

@rpkaul  I did not make any changes to the code that prevents people from joining the wrong team, specifically: https://github.com/splewis/get5/blob/0.9/scripting/get5/teamlogic.sp#L30-L91

Are you sure this was working correctly in the previous versions? We're looking into the pause-system.

rsalus:
Nice work. I am migrating a CSGO server to my local bare-metal cluster running k3s/docker. This will be one of the first plugins I will be testing -- will let you know how it goes. In the meantime I'll take a step through the code and see if anything jumps out.

PhlexPlexico:
One thing I noticed when testing as well is that the version is still outputting 0.8.0-dev, may wanna update that as well?

nickdnk:
> One thing I noticed when testing as well is that the version is still outputting 0.8.0-dev, may wanna update that as well?

Fixed.

scripting/include/get5.inc
```diff
@@ -15,16 +35,1307 @@ enum Get5State {
...
-#define MATCHTEAM_COUNT 4
-
-enum MatchTeam {
-  MatchTeam_Team1,
-  MatchTeam_Team2,
-  MatchTeam_TeamSpec,
-  MatchTeam_TeamNone,
-  MatchTeam_Count,
+enum PauseType {
+  PauseType_Tactical = 1,  // Tactical Pause
```
splewis, line 39: Any particular reason to start at 1 here?

scripting/include/get5.inc
```diff
@@ -15,16 +35,1307 @@ enum Get5State {
...
+
+    public bool IsEqualToPlayerId(const char[] id) {
+
+        char ownId[64];
+        this.GetSteamId(ownId, sizeof(ownId));
+
+        return StrEqual(ownId, id);
+
+    }
+
```
splewis, line 79: There's a ton of superfluous extra space here that harms readability.

splewis:
I did one pass, and didn't see any real issues. I do intend to do some manual testing before merging this, though just with 2 players.

Can you resolve the merge conflicts before I do the manual testing? We can also wait for #749 to go in, but I have no real preference for what is merged first.

nickdnk:
@splewis Updated so it compiles. Please take a look at the changes made to the pausing system as that's where most of the conflicts came from. The only thing I changed was adding two utility functions that would log out to the event system when the game pauses or unpauses, but this was not adjusted for Phlex' new tech pause fixes.

scripting/get5/pausing.sp
```diff
@@ -2,6 +2,38 @@ public bool Pauseable() {
...
+
+    LogDebug("Calling Get5_OnMatchPaused()");
+
+    Call_StartForward(g_OnMatchPaused);
+    Call_PushCell(event);
+    Call_Finish();
+
+    EventLogger_LogAndDeleteEvent(event);
+
+    return Pause(g_FixedPauseTimeCvar.IntValue, MatchTeamToCSTeam(team));
```
PhlexPlexico, line 17: This needs to be changed I think - if a user has a set tactical pause time but unlimited tech pause time, will this still hit? We had a user in Discord report that their technical pauses are still being fired off with timers. May need an if statement to check for what kind of pause it is and if those values are non-zero?

tapir:
I've been testing this branch together with some of the other PRs for the last 2 weeks. Looks like stable enough. Just FYI

nickdnk:
I've managed to get the events properly documented using OpenAPI/Swagger and the mkdocs plugin called `swagger-ui-tag` - this really helps.

I've had to add a few more objects, such as `Get5Winner` to represent a team + side for winning actions (map, series, round) and `Get5Weapon` to represent a weapon (name + id). It's a lot easier to document things when we can wrap them in objects and reuse them. This is of course a breaking change for anyone using forwards, but I'm okay with that as we will make a complete changelog when this is 100% ready.

@PhlexPlexico If you clone this and run `mkdocs build`, you can see how it looks under Events now. I only need to add a few missing events (such as player connect/disonnect, knife round start etc.) and then it's done.

nickdnk:
@splewis I would like to rename the `MatchTeam` enum to `Get5Team` and the new `Side` enum to `Get5Side`. I imagine you don't care about `Side` since I added that myself, but what about `MatchTeam`?

The reasoning here is that I constantly get confused about CSTeam vs. MatchTeam. I think their naming does not really indicate which is which. Also, I have prefaced all the the new "classes" with `Get5`, such as `Get5Player` and `Get5Weapon`. I think it would make sense to follow this convention in every place we can. It also makes documentation more 1:1 when it comes to naming of objects.

nickdnk:
@PhlexPlexico @splewis @deLuXe83 @rpkaul @tapir @rsalus 

0.9 is ready for merging now. I have made all the adjustments to it I would like, and I believe everything in it is now working as intended (not including what was broken before and not fixed in this PR, obviously).

Please give the latest build of this a test: https://github.com/splewis/get5/suites/7387731463/artifacts/300626706 and let me know if you have any final feedback.

scripting/include/get5.inc
```diff
@@ -15,16 +41,1577 @@ enum Get5State {
...
-  MatchTeam_Count,
+enum Get5PauseType {
+  Get5PauseType_None,      // Not paused
+  Get5PauseType_Tech,      // Technical pause
+  Get5PauseType_Tactical,  // Tactical Pause
+  Get5PauseType_Admin,     // Admin/RCON Pause
+  Get5PauseType_Backup     // Special type for match pausing during backups.
 };
 
+methodmap Get5StatusTeam < JSON_Object {
```
splewis, line 52: I find the spacing  here a bit excessive, especially since this is going to be read by any other developer interfacing with this.

Can we remove the blank lines between the class & first method, and the blank link after the last method? 

TBH, I'd be willing to drop the lines between individual methods/properties, but I don't have a very strong opinion. They're just boilerplate, so I would probably lean toward removing the lines.

scripting/get5/pausing.sp
```diff
@@ -2,31 +2,55 @@ public bool Pauseable() {
   return g_GameState >= Get5State_KnifeRound && g_PausingEnabledCvar.BoolValue;
 }
 
+public bool PauseGame(MatchTeam team, Get5PauseType type, int pausesLeft) {
+
```
splewis, line 6: Please remove this extraneous line.

(My opinion is basically that ofhttps://google.github.io/styleguide/cppguide.html#Vertical_Whitespace - this does not aid readability)

scripting/get5/util.sp
```diff
@@ -745,3 +815,37 @@ public bool IsJSONPath(const char[] path) {
...
+  return RoundToFloor((GetEngineTime() - timestamp) * 1000);
+}
+
+public int GetRoundsPlayed() {
+  return GameRules_GetProp("m_totalRoundsPlayed");
+}
+
+// Not entirely sure how this works, but it does work.
+// Also tested on Nuke with bombsites right on top of each other.
+stock Get5BombSite GetNearestBombsite(int client)
```
splewis, line 829: Bracing should be on the prior line for consistency.

scripting/get5/util.sp
```diff
@@ -745,3 +815,37 @@ public bool IsJSONPath(const char[] path) {
...
+{
+	int playerResource = GetPlayerResourceEntity();
+
+	if (playerResource == INVALID_ENT_REFERENCE)
+	{
+		return Get5BombSite_Unknown;
```
splewis, line 835: Bracing is inconsistent.

scripting/include/get5.inc
```diff
@@ -15,16 +41,1577 @@ enum Get5State {
...
+        self.SetMatchId(matchId);
+        self.SetTeam1Name(team1Name);
+        self.SetTeam2Name(team2Name);
+        return self;
+    }
+}
+
+methodmap Get5BackupRestoredEvent < Get5MapEvent {
+
+    public bool SetFileName(const char[] file)
```
splewis, line 932: Inconsistent brace

scripting/include/get5.inc
```diff
@@ -15,16 +41,1577 @@ enum Get5State {
...
+        return self;
+    }
+}
+
+// All other events
+
+methodmap Get5PlayerSayEvent < Get5PlayerTimedRoundEvent {
+
+  public bool SetCommand(const char[] command)
+  {
```
splewis, line 1038: Inconsistent brace

deLuXe83:
Hi guys.
thanks for finishing Version 0.9 :) (Y)

I just tested it, and had the following errors:
`get5_stats_players`, CONSTRAINT `get5_stats_players_matchid` FOREIGN KEY (`matchid`) REFERENCES `get5_stats_matches` (`matchid`)) `

any ideas?

nickdnk:
Your schema is probably incorrect. Please check it against the template in the repo.

Also, please post the logs surrounding the error with `get5_debug 1` set.

nickdnk:
Also, 0.9 is not done yet. Splewis said he would review the last PRs tonight. We will release an official build then with a migration guide and all the changes.

deLuXe83:
yes i know :)

here is another error:
`[Get5] The map will change once the GOTV broadcast has ended. `

keeps coming before mapchanging, and then mapchanging is not working :)

PhlexPlexico:
How long is your gotv delay? That would be how long you have to wait, plus another 15 seconds or so?

deLuXe83:
tv_dealy is 15

not changed to "old version"... there it worked smoothly

deLuXe83:
> Your schema is probably incorrect. Please check it against the template in the repo.
> 
> Also, please post the logs surrounding the error with `get5_debug 1` set.

has there bbeen changes in the SQL syntaxes?

nickdnk:
From 0.7.2 and 0.9, yes. I know it works fine with the schema and setup on the branch, so there is something not right with your database. Please post the logs surrounding the mysql error with `get5_debug 1` set, so we can see what it's doing.

deLuXe83:
okay thx, i'll look into it...
i remember i once updated the scheme a few month ago...

nickdnk:
Also please make sure you have updated your translations, as it will throw exceptions if you didn't.

nickdnk:
Also, please, as I said before, post logs with `get5_debug 1`.

deLuXe83:
ok, i at least think my SQL scheme shoould be correct... 
i now deleted it and re-created it, with the sql dump.
I had a blank space " " in my cfg file:
`"Match"
{
	"matchid"		"Match 5"
	"match_title"	"Match 5"
	"num_maps"		"2" // Must be an odd number or 2. 1->Bo1, 2->Bo2, 3->Bo3, etc.`
 
maybe this caused it...

but had that on earlier versions as well....

deLuXe83:
i will now leave it blank in my cfg... as mentioned in the docs ;)

`It is strongly recommended that you always leave the matchid blank, as MySQL will then manage the IDs for you.`

nickdnk:
In the previous version, a separate parameter existed to override the match ID, which was confusing. You must now leave it blank, as you found out. That should solve the problems.

PR #749, opened 2022-04-14
Include timeouts for tech pauses.

PhlexPlexico:
This PR includes the following two new config options and features:

- `get5_tech_pause_time "0"` - If set to a positive non-zero value, this will give a time limit for tech pauses before any team can unpause. Includes a new message when 30 seconds remain on a tech pause, and when the grace period expires.
- `get5_max_tech_pauses "0"` - If set to a positive non-zero value, this will also limit the amount of tech pauses a team may have during freeze time.

These features were requested in the [Discord](https://discord.gg/zmqEa4keCk) as there were some TOs mentioning that users were abusing the tech pauses, and the way ESEA handles it is with using timers for when users can unpause, and limits on how many tech pauses are allowed. One stop-gap to prevent teams abusing the timer is that once the team who has called the tech pause chooses to resume, their respective timers will stop as well. Up until the grace period expires, both teams must agree to resume the match. The max tech pauses do not increase until the grace period for that specific pause has run out.

Debug statements have been added as well to include them in team variables, such as the time remaining in a tech pause, how many tech pauses a team has used, and which team has called a tech pause.

Included with this is three new phrases that should be translated to other languages, but I believe it's not required as there are some translations already missing.

There is also a bug fix included that allowed users to unpause tech calls without the acceptance of the other team, once one tech pause had been called. I'm not sure if this was initially intended as a means of avoiding people abusing tech pause/unpause, but I believe this feature may be the next best thing.

I've tested this to the best of my ability, but I'm certainly not in QA so there may be some extreme edge cases that I have missed.

### DIFF
--- a/translations/get5.phrases.txt
+++ b/translations/get5.phrases.txt
@@ -143,7 +143,7 @@
     "TechPausePausesRemaining"
     {
         "#format"       "{1:s},{2:d}"
-        "en"            "Team {1} has {2} technical pauses remaining."
+        "en"            "Team {1} has {2} technical pause(s) remaining."
     }
     "MatchUnpauseInfoMessage"
     {
--- a/translations/get5.phrases.txt
+++ b/translations/get5.phrases.txt
@@ -133,7 +133,7 @@
     "TechPauseTimeRemainingMinutes"
     {
         "#format"       "{1:d}"
-        "en"            "{1} minutes remaining of tech pause time before anyone can resume."
+        "en"            "{1} minute(s) remaining of tech pause time before anyone can resume."
     }
     "TechPauseNoTimeRemaining"
     {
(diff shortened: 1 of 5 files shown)


translations/get5.phrases.txt
```diff
@@ -140,6 +140,11 @@
         "#format"       "{1:s}"
         "en"            "Sorry, team {1} has no more tech pause time. Please use tactical pauses."
     }
+    "TechPausePausesRemaining"
+    {
+        "#format"       "{1:s},{2:d}"
+        "en"            "Team {1} has {2} technical pauses remaining."
```
nickdnk, line 146: I would suggest `Team {1} has {2} technical pause(s) remaining.` so it doesn't look silly if they have 1 pause remaining.

PhlexPlexico:
So there was some issues with the .stop command, but it seems to be not in any relation to this as it has happened prior to this. I think we should be good to merge this in :)

Thanks!

splewis:
Looks good to me. The default unlimited tech pauses was always a precarious situation :)

Issue #586, opened 2020-10-26
Suggest using gitbook.com has a documentation management tool


ericstgo87:
Hi guys, 

I've seen it's quite a struggle to maintain the documentation using the github Wiki functionality as it is not possible to send pull request. 
There is a solution to this issue, it would be to create a separate github repository containing all the documentation and then link the repository to the gitbook.com free solution. 

Hope this makes sense ;)

Have a good day

nickdnk:
I just discussed this with splewis and agree that a better solution for documentation is required.

Drarig29:
Mkdocs could be a nice idea too, in this repo or another one (maybe `get5-docs`). This can be hosted on GitHub Pages.

Here is a documentation I'm making right now (it's still under construction). It uses the Material theme for mkdocs: https://drarig29.github.io/brackets-docs/

nickdnk:
@ericstgo87 Couldn't we have the docs be part of *this* repo and just set up a GitHub webhook to deploy it on some free service? I would prefer that PRs have documentation changes in them.

PR #752, opened 2022-05-19
Create Github Pages Documentation

PhlexPlexico:
Hi,

I've gone and moved the documentation from the wiki over to Mkdocs using GitHub pages. I've amalgamated a good chunk of the documentation both from what is in the wiki, the README, and what was also in the code and not present in both (i.e. pausing/freezing players during vetoes, pause events, etc.). 

In order for this to work, the new GitHub action will create a new branch called `gh-pages` which can be seen [here](https://github.com/PhlexPlexico/get5/tree/gh-pages). It compiles all the markdown pages located in the new documentation directory and creates a nice website to with the material theme to navigate through. I figured it may be easier to move documentation to a more maintainable solution (per #586), where anyone can edit the documentation.

Right now, the only pages I have omitted are the following:
- [Match backups](https://github.com/splewis/get5/wiki/Match-backups) as it is described how to restore match backups in the usage section for the commands.
- [get5 web panel and the get5_apistats plugin](https://github.com/splewis/get5/wiki/Match-backups) since it is also marked that the web panel itself is unsupported, and should not be used.
- [Recording match demos](https://github.com/splewis/get5/wiki/Recording-match-demos) since I feel most people shouldn't need to find how to do this since it is a bit out of scope for the project.
- [get5 vs eBot](https://github.com/splewis/get5/wiki/get5-vs-eBot) was removed as well, since we don't really need justification on which to use, if a person is already in this repository.

I can also reinstate these pages as well, if you'd like, but I figured I would try and keep the documentation closer to "usage of this plugin" instead of "other things that you can do in conjunction with this plugin".

Oh, and I've also updated the example match JSON to include the latest map pools since that was skipped, updated the workflow for the build to allow for one multi-line run instead of 20 lines of `run:`, and updated the README to point to the new docs site once it gets enabled.

The only thing that would need to be done if this gets merged is to go to Settings -> Pages and enable the source on the `gh-pages` branch once built.
image

Once complete, you will be able to see the documentation website, like you can see [here as an example.](https://phlexplexico.github.io/get5/).

One possible TODO before this goes in as well, would be to create a PR template that has some acknowledgement boxes such as "I have updated the documentation if there were any new changes". For example, if someone creates a new Event or config option, they will have to include it in the events documentation, so it doesn't get lost in code. If all looks good, we can take this out of draft and merge it in.

Thanks!


splewis:
Looks good! Let's check it out ;)

Issue #699, opened 2021-09-05
[BUG] First round restore swap issues


TandelK:
### Expected behavior
When using stop command on the very first Round for Technical reasons and restoring round00.cfg backup file it should just reset the round itself properly back to their teams. 

### Actual behavior
When Using backup00.cfg with the team which has swapped after knife  for side selection, it resets the  sides again to their old team.

### Steps to reproduce
Plugin version: 0.7.3-dev
Sourcemod Version = 1.10.0.6509

Steps to reproduce 
1)  Create Match with Knife for Side Selection
2) Start Knife round and then swap the sides. 
3) In first round use the stop  command and check the  side of the team that swapped has been resetted.

TandelK:
@splewis Does the Round Restore actually use Official Valve Backup ? If so it is not getting updated after the teams have been swap. I checked the backup created by get5 which actually while creating first time does properly stores the Team Swaps but as soon as we load the backup of round00 it seems that it is getting swapped because it might be using Valve Official backup and it does not seems to have updated after Knife Round.

arildboifot:
Might be related to #652 ?

PhlexPlexico:
Okay, so I think I'm going to use this issue to consolidate a few bugs in relation to the backup system if that's alright. And some possible solutions/patches that can be made here.

### Side Restore Issues

@TandelK had sent me some backup files to help debug this situation in this ticket. I believe that every pre-round start, the backup system writes a backup file, as seen here - 
https://github.com/splewis/get5/blob/5afc25b5f27e2bae923c2e47e3cfbc901259bbc6/scripting/get5.sp#L1150-L1166

However, when the knife round ends this function I believe technically runs in while we're still selecting sides, and the teams are still in their initial configuration that was sent to the match server. So, when a team calls `!swap`, the prelive value is still set to `team1` being on `team1_side` and `team2` being on `team2_side`. We can see that during the `PerformSideSwap` call, we don't call any updates to the backup structure. This can be seen in this call, that no calls to writing the backup is performed.

https://github.com/splewis/get5/blob/5afc25b5f27e2bae923c2e47e3cfbc901259bbc6/scripting/get5/kniferounds.sp#L26

[...]

PR #753, opened 2022-05-22
Change gMapSides when knife round swaps.

PhlexPlexico:
This should fix Side Restore Issues in #699.

Currently on round 0, it appears that the temp valve backup is not created, which causes us to hit the conditional in `RestoreGet5Backup()` and calls `SetStartingTeams()` in `teamlogic.sp` on Line 301. So instead of rewriting a bunch of logic, we just set the current `g_MapSides` value to be whichever side choice is selected by the `!swap` command.

### DIFF
--- a/scripting/get5/kniferounds.sp
+++ b/scripting/get5/kniferounds.sp
@@ -42,6 +42,13 @@ static void PerformSideSwap(bool swap) {
         }
       }
     }
+    // Make sure g_MapSides has the correct values as well,
+    // that way set starting teams won't swap on round 0,
+    // since a temp valve backup does not exist.
+    if (g_TeamSide[MatchTeam_Team1] == CS_TEAM_CT)
+      g_MapSides.Set(GetMapNumber(), SideChoice_Team1CT);
+    else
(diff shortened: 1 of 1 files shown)


tapir:
I get 

```
ERROR: Failed to compile get5, from ./smbuild
./scripting/get5/kniferounds.sp(48) : error 017: undefined symbol "GetMapNumber"
./scripting/get5/kniferounds.sp(50) : error 017: undefined symbol "GetMapNumber"
```

any ideas?

PhlexPlexico:
Make sure you don't have any other changes, I've never used smbuild as I've used my own Environment.

https://github.com/PhlexPlexico/get5/suites/6611108015/artifacts/248719183

But it looks like it compiled successfully here, so you can use that if you'd like.

tapir:
Hmm I have the 0.9 changes pulled too. Probably why it happens. Any chance to get this rebased to 0.9 since from what I understand it will be the next release

nickdnk:
@tapir We're more likely to rebase 0.9 on this though. I will do that once this gets merged in.

tapir:
Thanks a lot. Turns out was just a matter of renaming it to Get5_GetMapNumber() btw.

Issue #471, opened 2020-02-23
Coach moves to spectator after knife round swap


slawteruk:
If making a feature request, you should delete all the pre-filled text here.
If reporting a bug, fill in the following sections:

### Expected behavior
When the Knife round is finished and the winning team hits "!swap", each teams coach is swapped teams but stays in coach mode

### Actual behavior
When the Knife round is finished and the winning team hits "!swap", each teams coach is swapped teams, but are then moved into spectators. The entire time sv_coaching_enabled is correctly set to 1

### Steps to reproduce
Join server with two players and two coaches. Ready up and complete the knife round. Winner of knife round should type the !swap command. Once swapped, both coaches will now be spectators and will need to rejoin the server if they wish to become a coach again. 

Please note that "latest" is NOT a version! See https://github.com/splewis/get5/wiki/Debugging-help#finding-get5-version if you need help finding these. Adding the output file produced by ``get5_debuginfo`` is extremely helpful.

- Plugin version:
- Sourcemod version:
- Steps to reproduce (please be specific):

splewis:
Can you fill out: 

> Plugin version:
Sourcemod version:
Steps to reproduce (please be specific):

And:

Do you have a match config that can be used to reproduce this?

slawteruk:
Get5 Version: 0.7.1
SM Version: 1.10.0 build 6462
MM Version: 1.10 build 971

Steps:
1. Join server with 1 person on each team and 1 person coaching each team (Total 4 connections)
2. Both T and CT ready up and then use !forceready to start - 2 coaches are currently in "coach mode"
3. Knife round begins - 2 coaches still in "coach mode" 
4. End of the knife round, winner selects to swap sides using !swap command - Both coaches are then swapped to spectators 
5. Both coaches then have to leave the server entirely, rejoin and then re-enter the coach command. 

Our config is set up specifically for our API, so I can't put it on here publically. If you have a contact email or a way for us to communicate privately, I'll happily guide you through the process so you can see it for yourself. 

I can be contacted at michael@gameye.com if that helps you.

ShibaeO:
hi thx for your tip, but in config should i add coach steam id in spec or in the team ?
thx for your awnser

TandelK:
Coach Steam IDs have to be added in Teams and when the Coach joins the server he or she has to type !coach to join coach role.

ShibaeO:
thx for your awnser :)

kubo6472:
[...]

PR #754, opened 2022-05-26
Match Restore Coach Fix

PhlexPlexico:
Per #699 there is an issue during match restores either using `get5_loadbackup` or `!stop` in chat. Once the match is restored, a coach is dropped onto the team, and unable to `.coach` back. Maybe there was a change to the coaching mechanism in CS:GO, but this fix simply allows users to use `.coach` In freeze time, as was intended by the looks of it from the comments?

Thanks!

### DIFF
--- a/configs/get5/example_match.json
+++ b/configs/get5/example_match.json
@@ -1,7 +1,7 @@
 {
 	"matchid": "example_match",
 	"num_maps": 3,
-	"players_per_team": 1,
+	"players_per_team": 5,
 	"coaches_per_team": 2,
 	"min_players_to_ready": 1,
 	"min_spectators_to_ready": 0,
--- a/configs/get5/example_match.json
+++ b/configs/get5/example_match.json
@@ -2,6 +2,7 @@
 	"matchid": "example_match",
 	"num_maps": 3,
 	"players_per_team": 1,
+	"coaches_per_team": 2,
 	"min_players_to_ready": 1,
 	"min_spectators_to_ready": 0,
 	"skip_veto": false,
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -102,6 +102,7 @@ bool g_BO2Match = false;
 char g_MatchID[MATCH_ID_LENGTH];
 ArrayList g_MapPoolList = null;
 ArrayList g_TeamAuths[MATCHTEAM_COUNT];
+ArrayList g_TeamCoaches[MATCHTEAM_COUNT];
 StringMap g_PlayerNames;
 char g_TeamNames[MATCHTEAM_COUNT][MAX_CVAR_LENGTH];
 char g_TeamTags[MATCHTEAM_COUNT][MAX_CVAR_LENGTH];
@@ -113,6 +114,7 @@ char g_MatchTitle[MAX_CVAR_LENGTH];
 int g_FavoredTeamPercentage = 0;
 char g_FavoredTeamText[MAX_CVAR_LENGTH];
 int g_PlayersPerTeam = 5;
+int g_CoachesPerTeam = 2;
 int g_MinPlayersToReady = 1;
 int g_MinSpectatorsToReady = 0;
 bool g_SkipVeto = false;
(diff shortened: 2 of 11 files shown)


PhlexPlexico:
Moving this to a draft because I would like to make further improvements on how coaches are handled. Mainly the following:

- Use a global ArrayList to keep track of which coach is on which team, that way when a player disconnects and reconnects, they can immediately be thrown back onto the coaching team.
- Add the team coaches to the backup structure, that way during a match restore they will be placed onto the correct coach slot for teams as well.

PhlexPlexico:
So player placement on round restore gets to be reaaaaal buggy. However, I think we may be best off just blocking on who can coach on a round restore as well. I'd like to get some more people to test these new features and let me know how it goes, but I don't think I ran into anything too terrible when working with the backup structure.

scripting/get5.sp
```diff
@@ -107,6 +107,7 @@ char g_FormattedTeamNames[MATCHTEAM_COUNT][MAX_CVAR_LENGTH];
 char g_TeamFlags[MATCHTEAM_COUNT][MAX_CVAR_LENGTH];
 char g_TeamLogos[MATCHTEAM_COUNT][MAX_CVAR_LENGTH];
 char g_TeamMatchTexts[MATCHTEAM_COUNT][MAX_CVAR_LENGTH];
+char g_TeamCoaches[MATCHTEAM_COUNT][64];
```
splewis, line 110: I believe it's possible for a team to have multiple coaches (i.e. coach + analyst) - I haven't verified that anytime lately though.

PhlexPlexico: Yeah I think this still may be the case, I'll fix this and update it to be an array in coming days!

PhlexPlexico:
Okay, to address one of the coach bug issues someone was intermittently having as well, I've introduced `coaches_per_team` in the match config. It's entirely optional, but by default it only allows two coaches per team. 

I've also updated the structs to be plural, since we can allow multiple coaches, and they both load from JSON and KeyValue files. Documentation has been updated to show for it as well.

`g_TeamCoaches` now acts like the player array does, and adds based on if it exists or not. I don't think this should be merged just yet, but everything should be working as intended again. @kubo6472 would you mind testing these changes out if you have the time? I know you were asking about the add coach feature and this should be in there as well.

configs/get5/example_match.json
```diff
@@ -2,6 +2,7 @@
 	"matchid": "example_match",
 	"num_maps": 3,
 	"players_per_team": 1,
```
nickdnk, line 4: Might as well make this 5, no?

nickdnk:
Fixes https://github.com/splewis/get5/issues/471

scripting/get5/teamlogic.sp
```diff
@@ -354,6 +424,23 @@ public bool AddPlayerToTeam(const char[] auth, MatchTeam team, const char[] name
   }
 }
 
+public bool AddCoachToTeam(const char[] auth, MatchTeam team, const char[] name) {
+  char steam64[AUTH_LENGTH];
```
nickdnk, line 428: Move this below the `(team == MatchTeam_TeamSpec)` check.

scripting/get5/teamlogic.sp
```diff
@@ -223,6 +283,24 @@ public MatchTeam GetAuthMatchTeam(const char[] steam64) {
   return MatchTeam_TeamNone;
 }
 
+public MatchTeam GetAuthMatchTeamCoach(const char[] steam64) {
+  if (g_GameState == Get5State_None) {
+    return MatchTeam_TeamNone;
+  }
+
+  if (g_InScrimMode) {
+    return IsAuthOnTeamCoach(steam64, MatchTeam_Team1) ? MatchTeam_Team1 : MatchTeam_Team2;
```
nickdnk, line 292: Wouldn't it be safer here to not assume that false == team2 ?

scripting/get5/teamlogic.sp
```diff
@@ -74,21 +85,39 @@ public Action Command_JoinTeam(int client, const char[] command, int argc) {
...
     return Plugin_Stop;
   }
 
   return Plugin_Stop;
 }
 
+public bool CheckIfClientCoaching(int client, MatchTeam team) {
+    // Force user to join the coach if specified by config or reconnect.
+    char clientAuth64[AUTH_LENGTH];
+    GetAuth(client, clientAuth64, AUTH_LENGTH);
```
nickdnk, line 112: Move getAuth inside the "if coaching" check. No reason to fetch if it should not be evaluated.

nickdnk: or just return false early at the beginning if g_CoachingEnabledCvar.BoolValue is false

PR #765, opened 2022-06-30
Implement "random" value for "veto_first"

tapir:


PhlexPlexico (commented):
LGTM - So long as this was tested, shouldn't be an issue.

https://github.com/splewis/get5/blob/422f745270d854023608773cad14bd19120f205d/scripting/include/get5.inc#L20-L26

It also looks like 0 and 1 matches the enum, so I have no reservations for this!

PR #767, opened 2022-07-02
Bring in check to not allow users the ability to unpause admin pauses.

### LINKED ISSUE #737: [FR] Stop players' ability to un-pause the game when paused by admin
When the match is paused by an admin, players shouldn't be able to un-pause the game themselves.

PhlexPlexico:
Closes #737

Largely untested, so would need someone to confirm this works as intended, but I believe it should be default be false and the only way to really trigger it would be console pausing.


nickdnk:
I would recommend putting the variable reset inside the Unpause() helper method to ensure it always gets set to false, also if we move code around down the line.

nickdnk:
@PhlexPlexico Current PR does not reset the variable at all. I think you forgot to add it to the utility function?

1. Reset it to false when the game is paused by **non-admin** (so that it gets reset even if there is a pause followed by a changelevel and hence no call to `Unpause`).
2. Reset it to false in the `Unpause` utility function.

PhlexPlexico:
Hey sorry about that, guess I missed saving the util file when committing my changes to git. It should be good to go now, I've included back the setting `g_PausedByAdmin` to false after the client checks, and made sure the util class was actually included this time with the proper change. Lemme know if I've missed anything else, sorry about that!

PhlexPlexico:
image


Looks like this is working as intended. I think we should be good to merge this in. One was paused via RCON to the server, and then once by users.

nickdnk (approved):

PR #773, opened 2022-07-15
CI Integration For Releases/Nightlies

PhlexPlexico:
Update build script to build release tags.

Update issue template to remove wiki link.

Update documentation to include links to proper releases to the release/tag on GitHub.

Update Actions to run on more recent version of Ubuntu for compiling to ensure it can still compile on newer systems.

Here is an example as to what the pre-release looks like https://github.com/PhlexPlexico/get5/releases/tag/0.8.0-909bb3 as well as the tag here - https://github.com/PhlexPlexico/get5/releases/tag/0.8.0-909bb3


.github/workflows/build.yml
```diff
@@ -1,15 +1,20 @@
...
   build:
-    runs-on: ubuntu-18.04
+    name: Compile & Generate Tagged Nightly
+    runs-on: ubuntu-22.04
     env:
       OUTPUT_PATH: output/get5
       OUTPUT_SM_PATH: output/get5/addons/sourcemod
       SM_JSON_INC_PATH: get5/dependencies/sm-json/addons/sourcemod/scripting/include
       STEAMWORKS_URL: https://raw.githubusercontent.com/KyleSanderson/SteamWorks/1.2.3c/Pawn/includes/SteamWorks.inc
+      GET5_VERSION: 0.8.0
```
nickdnk, line 17: Shouldn't this be read from commit tag?

.github/workflows/build.yml
```diff
@@ -1,15 +1,20 @@
-name: Build
+name: Build and tag nightlies
 
-on: [push]
+on:
+  push:
+    branches:
+      - master
```
nickdnk, line 6: And only on tag here, no?

.github/workflows/build.yml
PhlexPlexico: I kind of threw it in here because this would also avoid a headache of trying to regex out the version number. Maybe we should put it in here and place that in the `version.sp` file instead?

.github/workflows/build.yml
PhlexPlexico: I'm still kind of working on this... I think on push to dev we'll create the tag and a pre-release, and then we can have a check on master to create a tag and draft up a full release. 

We could also have dev just create tags only, with no release, and point users to the tags as a "bleeding edge" link, with new features from the dev branch?

.github/workflows/build.yml
nickdnk: So you want to have this document be the source of truth for the version, not the tags?

PhlexPlexico: I was thinking possibly use this as a single point for the version number, yes.

Right now the only place that the true version exists is in the `version.sp` which is a 4 line file and that was updated to point to new version releases. This is essentially just moving it to another file for further automation, where nightly tags (which would build, and attach the `.zip` and `.tar.gz` files to the tag) can be generated on each push to the dev branch.

I feel this has benefits because then if a build fails when it's pushed to dev, a tag will not be created for that nightly or built until it is resolved. For tagging nightlies, I was thinking we could have the tags as something like `v0.8.0-{short_commit_hash}`, and for full stable releases, we can automate a draft release that will just be `v0.9.0` (or whichever version number we're at when we are stable), when pushed to master. I still need to workout the logic for this, but I was thinking this may be a better solution for CI purposes while being able to keep a bleeding edge nightly that can exist for greater than 90 days like the current release, and have a way of drafting stable releases quickly. I think we can even scrap the "Pre-Release" section that GitHub has as well, as I believe tags can also have attachments, too.

PhlexPlexico:
Putting this into draft for now while I get some more things sorted with it!

PhlexPlexico:
Okay, per the conversation above I think it may be alright to move the version over into the release file itself? It will help us keep track and automate the task for changing the release versions in one file, as well as allowing the CI to change the commit sha as needed for whenever we push a nightly.

Now, whenever we push to the development branch, we create a tag that is `v${version}-{commit}` and when a full release is generated, a tag is created with only the version number. So when we are ready to release to stable, we just create and merge a PR from development -> master, and the CI takes care of the rest. Tagging, uploading, and generating a changelog. Right now, it puts the release in draft in case there is any edits to be made to the change log as it only takes the link to the commit, plus I believe the title of the commit/merge that was done. 

If we want a different way, I'm all ears but I think this may accomplish what we want for "bleeding edge" releases, and helping us draft up more stable releases in a faster fashion by including all the changes within a draft.

scripting/get5/version.sp
```diff
@@ -1,4 +1,4 @@
 #tryinclude "manual_version.sp"
```
nickdnk, line 1: What's this? `manual_version`?

PhlexPlexico: Me neither, I think it should be okay to remove it, I assume it was for development at some point?

splewis: It's used to stamp manual version numbers (including the build #) in my Jenkins instance. (ctrl+f for 
 manual_version @ https://ci.splewis.net/job/get5/561/console )

I think it would be reasonable to work toward not relying on the jenkins instance for get5 at all, but I'm not quite sure we're there yet: the link in documentation points to https://github.com/splewis/get5/actions, and I don't think it's very easy to find the "latest dev build" from that.

splewis: Another aside: the directory structure of the github build actions seems excessively nested: `get5.zip/output/get5/addons/sourcemod/...` - I would think just `get5.zip/addons/sourcemod/...` would be preferable, though I haven't checked into how it's configured.

nickdnk: @splewis We're planning on getting all of these things fixed as soon as we can merge in 0.9, including having a new official release + easier linking to dev builds. That's why we added the development branch, so we can have only stable releases and tags on master.

PhlexPlexico: > Another aside: the directory structure of the github build actions seems excessively nested: `get5.zip/output/get5/addons/sourcemod/...` - I would think just `get5.zip/addons/sourcemod/...` would be preferable, though I haven't checked into how it's configured.

This has been fixed in this CI update as well, where it's simply two attachments in a tag(or pre-release, per GitHub) `get5.zip` and `get5.tar.gz` which only contained the `README` and `LICENSE` at the root directory, followed by the directories `addons`, `cfg` and I believe the translations and whatnot are nested under `addons/sourcemod` so a user can simply unzip the file in their server instance.

As for the documentation, I was planning on linking to the tags since those will sort by most recent and if you click on a tag will show the attachments, and full releases to go to the release page, if that works?

scripting/get5/version.sp
```diff
@@ -1,4 +1,3 @@
-#tryinclude "manual_version.sp"
 #if !defined PLUGIN_VERSION
-#define PLUGIN_VERSION "0.8.0-dev"
+#define PLUGIN_VERSION "<VERSION>-<COMMIT_HASH>"
```
nickdnk, line 2: I would suggest we use something else as the placeholder, as developers will get literally `<VERSION>-<COMMIT_HASH>` as their version number if they compile outside of CI.

Maybe:
`#define PLUGIN_VERSION "<DEVELOPMENT>"`

And then in the CI logic, replace `<DEVELOPMENT>` with `x.y.z-aabbccdd` in one go, instead of two calls to `sed`? Should be possible.

PhlexPlexico: Sounds good!

splewis: Why the placeholder can't remain a meaningful "0.9.0-dev" or similar for someone checking this out & compiling it manually?

nickdnk: Well, 0.9 as a fixed version forever might also be slightly odd?

nickdnk: If you move the version back, you have the problem of not having a single source of truth for the version.

PhlexPlexico: How so? We can remove it from all the CI and use grep and see to grab for these values. I'm working on it right now and think I have a possible solution for this where the only version that exists will be in `version.sp` with that `tryinclude` still there until we can move away from Jenkins fully

nickdnk: Also totally fine, but I was under the impression you wanted the version in the CI file. I think it's better in the `version.sp` file as well and then regex it out for tagging. My point was just that I think we should only have one single place where we change it.

nickdnk: I think this approach is good.

PhlexPlexico: Depending on the amount of development and features going in, I think it would be nice to have something that is released fairly often, especially for testing new features for public use. We can only do so much testing as a group, and enabling people to just grab the latest and greatest and report bugs would be far better than catching them in a full release.

PhlexPlexico:
Okay so I think we may want to break the linting into a separate PR since I'm not sure if I'm going to have the time to get it done in a timely fashion, there's a bunch of bash logic that I'm going to have to work through since the python script forces git to mark them as "changed" even though nothing has changed.

I think this should be good to merged in, and nightlies can start being created on development. 

Per @splewis comment, I've improved the versioning to keep the version and `-dev` there. So whenever someone builds from source, we will know when a user builds direct from cloning vs just downloading a nightly.

The build outputs have been updated as well now to be more inline with the build process on Jenkins where the structure is like so:
```json
"get5.zip" {
  "LICENSE.md"
  "README",
  "addons" {
    "sourcemod" {
      "configs" {
        "...stuff"
        },
       "plugins" {
         "...stuff"
         },
         "scripting" {
           "...stuff"
          },
          "translations" {
            "..stuff"
           }
         }
       }
    "cfg" {
      "...stuff"
    }
```

so drag and unzip/untar is supported (and can be directly installed as a .zip by DatHost, for example).

I've left the manual_version tryincludes now as well, since we are still using your Jenkins instance, and it can still build there. But all of this should now hopefully replace that :)

I've also updated the README to include a link to the latest full release, and then a link to all pre-release tags which is where the nightlies will be contained.

.github/workflows/development.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    - uses: rumblefrog/setup-sp@master
+      with:
+        version: '1.10.x'
+
+    - name: Get shorthand commit.
+      id: vars
+      run: |
+        echo "::set-output name=sha_short::$(echo ${{ github.sha }} | cut -c1-6)"
+        echo "::set-output name=get5_version::$(cat get5/scripting/get5/version.sp | grep -Eo \"[0-9]*\.[0-9]\.[0-9]-[a-zA-z]{3}\")"
```
nickdnk, line 33: Regex doesn't match stuff like `0.12.12-dev`

Needs to be `[0-9]+\.[0-9]+\.[0-9]+-[a-zA-Z]+` - no reason to lock it into 3 symbols at the end either

.github/workflows/development.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    - run: | 
+        wget $STEAMWORKS_URL -P steamworks
+        mkdir -p $OUTPUT_SM_PATH/plugins/disabled
+        sed -i -r "s#[0-9]*\.[0-9]*\.[0-9]*-[a-zA-z]{3}#${{ steps.vars.outputs.get5_version }}-${{ steps.vars.outputs.sha_short }}#g" get5/scripting/get5/version.sp
```
nickdnk, line 38: Same here. Better, but still allows something like `.3.3-dev`. Intentional?

PhlexPlexico: Intentional as we should be sticking to semantic versioning. If there's an error in the version number, it will at least be shown immediately on tag creation? Hopefully something that would be caught way sooner. Also fixed this one up in build.yml as well.

nickdnk: I don't understand. You want it to actually match to an invalid syntax and not error?

PhlexPlexico: Actually wait, with the new regex match it will still match something like `.12.2-dev`, so it will show up on tag creation without error. I feel like this is something that should not stop build creation though, as it's only a versioning number?

.github/workflows/build.yml
PhlexPlexico: As talked in the Discord, this will draft up a full release for us, so we just have to verify the release notes before publishing :)

PR #775, opened 2022-07-18
Coach Fixes for JoinGame

PhlexPlexico:
Add in timer for a fake client call as `ChangeClientTeam` breaks the functionality of coaches being able to swap with the in-game `coach ct/t` commands that we tie into.


scripting/get5/teamlogic.sp
```diff
@@ -8,11 +8,23 @@ public Action Command_JoinGame(int client, const char[] command, int argc) {
     if (clientTeam == CS_TEAM_NONE || clientTeam == CS_TEAM_SPECTATOR) {
       return Plugin_Continue;
     }
-    ChangeClientTeam(client, clientTeam);
+    DataPack playerPack = CreateDataPack();
+    playerPack.WriteCell(client);
+    playerPack.WriteCell(clientTeam);
+    CreateTimer(0.1, Timer_PlacePlayerOnJoin, playerPack);
```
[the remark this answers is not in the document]
nickdnk: Also, add some comments here explaining why we're not using `ChangeClientTeam`, so someone doesn't "change it back cause it's cleaner".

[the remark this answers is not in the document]
PhlexPlexico: Done and done, hopefully the comment isn't too large either for the explanation.

scripting/get5/teamlogic.sp
```diff
@@ -8,11 +8,34 @@ public Action Command_JoinGame(int client, const char[] command, int argc) {
...
 }
 
+public Action Timer_PlacePlayerOnJoin(Handle data, DataPack pack) {
+  pack.Reset();
+  int client = GetClientOfUserId(pack.ReadCell());
+  int clientTeam = pack.ReadCell();
+  if (client) {
+    FakeClientCommand(client, "jointeam %d", clientTeam);
+  } else {
+    LogDebug("We are not able to find the client. Refusing to jointeam.");
```
nickdnk, line 33: Either remove logdebug or make it more clear where it's coming from. I.e. "Player disconnected before they could be automatically placed on a team."

nickdnk (changes requested):

nickdnk:
I have an idea. Why don't we just directly call `Command_JoinTeam` instead of having this logic kind of duplicate here?

https://github.com/splewis/get5/blob/8c5e46a51b72723b6712d3c7e288201fd13a91fb/scripting/get5/teamlogic.sp#L37

This function overrides to the correct team anyway, so we should be able to just:
```c
if (g_CheckAuthsCvar.IntValue > 0 && !g_PendingSideSwap) {
    int clientTeam = Get5TeamToCSTeam(GetClientMatchTeam(client));
    if (clientTeam == CS_TEAM_NONE || clientTeam == CS_TEAM_SPECTATOR) {
      return Plugin_Continue;
    }
    // We are required to put a FakeClientCommand on a timer, 
    // due to ChangeClientTeam having some undefined behaviour
    // when a coach attempts to call .coach/sm_coach
    // and have the in-game Valve coaching call not fire off.
    // In order to avoid this behaviour, we have to use a FakeClientCommand
    // within a few hundred ms to ensure the client is placed, and can
    // call .coach on warmup.
    CreateTimer(0.1, Timer_PlacePlayerOnJoin, client, TIMER_FLAG_NO_MAPCHANGE);
  }
  return Plugin_Continue;
```
And in the callback:
```c
public Action Timer_PlacePlayerOnJoin(Handle data, int client) {
  Command_JoinTeam(client, "jointeam", Get5TeamToCSTeam(GetClientMatchTeam(client)));
}
```

scripting/get5/teamlogic.sp
```diff
@@ -1,16 +1,26 @@
 public Action Command_JoinGame(int client, const char[] command, int argc) {
-  if (g_GameState == Get5State_None || !IsPlayer(client)) {
-    return Plugin_Continue;
+  // We are required to put a FakeClientCommand on a timer,
```
nickdnk, line 2: I think it's good now, just clean up this comment, as there is now no `FakeClientCommand` call.

Suggest:

1. Add a check for g_CheckAuthsCvar + gamestate == none here, so we don't start timers every time a user joins if Get5 is not doing any team logic anyway.
2. Change comment to something like:
```
In order to avoid duplication of team-join logic, we directly call the same handle that would be called
if the user selected any team after joining. Since Command_JoinTeam handles the actual joining using a
FakeClientCommand, we don't have to do any team-logic here and it won't matter what we pass to Command_JoinTeam. The only thing that's important is that the command argument is empty, as that avoids a call to GetCmdArg in that function.
```

scripting/get5/teamlogic.sp
```diff
@@ -1,16 +1,26 @@
...
-      return Plugin_Continue;
-    }
-    ChangeClientTeam(client, clientTeam);
+public Action Timer_PlacePlayerOnJoin(Handle data, int userId) {
+  int client = GetClientOfUserId(userId);
+  if (!client) {
+    return; // Client has disconnected.
   }
-  return Plugin_Continue;
+  // second argument can be empty string because it's not used in the function. It's only bound to jointeam.
```
nickdnk, line 20: And maybe remove this comment as it's redundant if you add the above comment.

nickdnk:
I like the buffer check, just add comment to it explaining why. I added three comments on the current code.

scripting/get5/teamlogic.sp
```diff
@@ -1,18 +1,25 @@
 public Action Command_JoinGame(int client, const char[] command, int argc) {
-  if (g_GameState == Get5State_None || !IsPlayer(client)) {
+  if (g_GameState == Get5State_None || !g_CheckAuthsCvar.BoolValue) {
```
nickdnk, line 2: Why remove `!IsPlayer(client)` ?

scripting/get5/teamlogic.sp
```diff
@@ -49,14 +56,19 @@ public Action Command_JoinTeam(int client, const char[] command, int argc) {
   }
 
   char arg[4];
-  GetCmdArg(1, arg, sizeof(arg));
-  int team_to = StringToInt(arg);
+  int team_to;
+  // This is required as it avoids an exception due to calling this function
+  // from Timer_PlacePlayerOnJoin, which gets called from Command_JoinGame.
+  if (!StrEqual("", command)) {
+    GetCmdArg(1, arg, sizeof(arg));
```
nickdnk, line 63: Put
```
char arg[4];
int team_to;
```
  inside the if-block

nickdnk (approved):

scripting/get5/teamlogic.sp
```diff
@@ -1,18 +1,25 @@
...
+  // if the user selected any team after joining. Since Command_JoinTeam handles the actual joining using a
+  // FakeClientCommand, we don't have to do any team-logic here and it won't matter what we pass to Command_JoinTeam.
+  // The only thing that's important is that the command argument is empty, as that avoids a call to GetCmdArg in that function.
+  if (!g_PendingSideSwap) {
+    CreateTimer(0.1, Timer_PlacePlayerOnJoin, GetClientUserId(client), TIMER_FLAG_NO_MAPCHANGE);
   }
   return Plugin_Continue;
 }
 
+public Action Timer_PlacePlayerOnJoin(Handle data, int userId) {
```
nickdnk, line 15: Change this function to this. Gets rid of the early return and renames the `data` argument to `timer` just because that's what it is.

```c
public Action Timer_PlacePlayerOnJoin(Handle timer, int userId) {
  int client = GetClientOfUserId(userId);
  if (client) { // Client might have disconnected between timer and callback.
    Command_JoinTeam(client, "", 1);
  }
}
```

scripting/get5/teamlogic.sp
```diff
@@ -1,18 +1,25 @@
 public Action Command_JoinGame(int client, const char[] command, int argc) {
-  if (g_GameState == Get5State_None || !IsPlayer(client)) {
+  if (g_GameState == Get5State_None || !g_CheckAuthsCvar.BoolValue || !IsPlayer(client)) {
```
nickdnk, line 2: Invert the if-logic to reduce to 1 return and 1 if-block.
```c
public Action Command_JoinGame(int client, const char[] command, int argc) {
  if (g_GameState != Get5State_None && g_CheckAuthsCvar.BoolValue && IsPlayer(client) && !g_PendingSideSwap) {
    // In order to avoid duplication of team-join logic, we directly call the same handle that would be called
    // if the user selected any team after joining. Since Command_JoinTeam handles the actual joining using a
    // FakeClientCommand, we don't have to do any team-logic here and it won't matter what we pass to Command_JoinTeam.
    // The only thing that's important is that the command argument is empty, as that avoids a call to GetCmdArg in that function.
    CreateTimer(0.1, Timer_PlacePlayerOnJoin, GetClientUserId(client), TIMER_FLAG_NO_MAPCHANGE);
  }
  return Plugin_Continue;
}
```

nickdnk (changes requested):
These are the last two things. Promise :D

PR #776, opened 2022-07-19
Use user ID in player object

nickdnk:
It seems it's smarter to use https://sm.alliedmods.net/new-api/clients/GetClientUserId and store this in grenade victims, than to use steam ID, as it is unique for all intents and purposes within the lifetime of a server (rolling over at 2^16). This means we don't have to allocate a new `JSON_Object` for each time a molotov deals damage, and it also means we don't need to compare players using their steam ID.

Untested, but should work. I'll test it soon and then merge.

### DIFF
--- a/scripting/include/get5.inc
+++ b/scripting/include/get5.inc
@@ -355,20 +355,18 @@ methodmap Get5Player < JSON_Object {
       }
     }
 
-    public bool IsEqualToPlayerId(const char[] id) {
-      char ownId[64];
-      this.GetSteamId(ownId, sizeof(ownId));
-      return StrEqual(ownId, id);
+    property int Id {
+      public get() {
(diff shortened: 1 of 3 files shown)


scripting/include/get5.inc
```diff
@@ -355,20 +355,18 @@ methodmap Get5Player < JSON_Object {
...
-      char playerId[64];
-      player.GetSteamId(playerId, sizeof(playerId));
-      return this.IsEqualToPlayerId(playerId);
+    property int Id {
+      public get() {
+        return this.GetInt("id");
+      }
+      public set(int id) {
+        this.SetInt("id", id);
+      }
```
splewis, line 364: I would suggest making this explicitly called `UserId`, or at least adding a comment what kind of "id" this is.

PR #778, opened 2022-07-20
Remove player should remove from coach

nickdnk:
Hey

I'm fairly certain we need to do this when coaches are now in their own auth array.

@PhlexPlexico Kind of your area.

### DIFF
--- a/scripting/get5/matchconfig.sp
+++ b/scripting/get5/matchconfig.sp
@@ -836,9 +836,9 @@ public Action Command_AddPlayer(int client, int args) {
     }
 
     if (AddPlayerToTeam(auth, team, name)) {
-      ReplyToCommand(client, "Successfully added player %s to team %s", auth, teamString);
+      ReplyToCommand(client, "Successfully added player %s to %s.", auth, teamString);
     } else {
-      ReplyToCommand(client, "Player %s is already on a match team.", auth);
+      ReplyToCommand(client, "Failed to add player %s to team %. They may already be on a team or you provided an invalid Steam ID.", auth, teamString);
     }
 
   } else {
@@ -857,7 +857,6 @@ public Action Command_AddCoach(int client, int args) {
   }
 
   char auth[AUTH_LENGTH];
-  char steam64[AUTH_LENGTH];
   char teamString[32];
   char name[MAX_NAME_LENGTH];
   if (args >= 2 && GetCmdArg(1, auth, sizeof(auth)) &&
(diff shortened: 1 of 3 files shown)


scripting/get5/matchconfig.sp
```diff
@@ -882,6 +882,7 @@ public Action Command_AddCoach(int client, int args) {
     }
 
     if (!ConvertAuthToSteam64(auth, steam64)) {
+      ReplyToCommand(client, "Failed to parse steam ID: %s", auth);
```
PhlexPlexico, line 885: This is already called in a LogError, do we want to tell users if we fail to parse steam IDs? If so, I would recommend doing `ConvertAuthToSteam64(auth, steam64, false)` to avoid double reports in the server console.

PhlexPlexico (commented):
Right, didn't really check this logic but yeah it would make sense that if we want to kick a player or a coach, it should be checking both ArrayLists.

Also, it looks like `SwapScrimTeamStatus` is specifically used for ringers... 
https://github.com/splewis/get5/blob/92e0fb9e968d7449fda196a116d7870642451298/scripting/get5/get5menu.sp#L73-L90

https://github.com/splewis/get5/blob/4460f01d40f5aecbb646b744879f58865218454f/scripting/get5/matchconfig.sp#L1068-L1087

I don't think you would ever call a ringer for a coach, so I think this should be fine to leave? I don't think I could ever see logic of someone coaching a team to immediately be swapped in... So I think it would be fairly inconsequential if we left it?

scripting/get5/matchconfig.sp
nickdnk: User cannot see error logs. It also would not make sense to continue with the rest of the function of the auth is invalid.

I think doing both is fine. Perhaps changing the Log to a LogDebug inside the auth converter?

nickdnk: Ah, it shows error log in console?

We can remove the Reply perhaps, but it needs to return if false.

nickdnk: I also was a bit quick here. I have now modified it so that a false return either `AddPlayerToTeam` or `AddCoachToTeam` prints out a more accurate response.

I also removed the check you commented on, since it's already being done inside the `AddCoachToTeam` itself. We don't need it outside of it.

PhlexPlexico (approved):
Perfect, looks good to me! Thanks for picking this up, completely flew past my radar!

nickdnk:
I don't see a single `ReplyToCommand` using translations? Also, I don't think any console commands should support translations. It's simply too much work. It should be confined to the stuff printed in chat in my opinion.

PhlexPlexico:
Ah wait, yeah nevermind I'm thinking of the Get5_MessageToAll. My bad!

PR #781, opened 2022-07-21
Disable console "coach" command to make coaching logic easier to handle

nickdnk:
I'm pretty sure we better do this. There is no reason to let the ingame coach command mess around with the `.coach/sm_coach` logic we already have.

@PhlexPlexico  ?

### DIFF
--- a/scripting/get5/teamlogic.sp
+++ b/scripting/get5/teamlogic.sp
@@ -176,15 +176,16 @@ public void MoveClientToCoach(int client) {
 }
 
 public Action Command_SmCoach(int client, int args) {
-  char auth[AUTH_LENGTH];
   if (g_GameState == Get5State_None) {
     return Plugin_Continue;
   }
 
   if (!g_CoachingEnabledCvar.BoolValue) {
-    return Plugin_Handled;
+    ReplyToCommand(client, "Cannot set coaches when sv_coaching_enabled is false.");
+    return Plugin_Stop;
   }
 
+  char auth[AUTH_LENGTH];
   GetAuth(client, auth, sizeof(auth));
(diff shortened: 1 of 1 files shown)


PhlexPlexico:
I think we need to leave it as is, since there are a few cases in which the coach command is overridden, such as when backups are restored. I don't think we can remove this without breaking a few things and having to go fix them later.

nickdnk:
Can you elaborate? The only thing this does is prevent users from manually typing the coach command. A FakeCommand doesn't trigger this, does it?

PhlexPlexico:
Mainly two calls prevent this from being completely replaced. With the current commit, there would be an infinite loop if you try and call `.coach` during warmup - 

https://github.com/splewis/get5/blob/8c5e46a51b72723b6712d3c7e288201fd13a91fb/scripting/get5/teamlogic.sp#L171-L172

FakeClientCommand does call it like a regular hook. I believe we only want to further enhance the coaching system that is in place, and not completely replace it. If a user wants to call `coach ct` in console because that's what they're used to, they should still be able to do that and since the base game has support for it, we should do as few things as possible (check halftime phase, if coaching is enabled, write backup).

If we remove it and keep it like it is, I'm fairly certain there will be an exception on placing a coach because it will now recursively call `Command_Coach`.

nickdnk:
> Edit: Of course the code collapsing makes it look like all one function - however right now it doesn't appear it would, but I would still want full testing done with coaches to ensure it's working as intended, because I still don't like the thought of completely taking over the coaching system with our own.

It seems we're actually already doing this as the method just returns `Plugin_Stop` after redirecting the command. So I think this is working as intended.

Issue #647, opened 2021-03-04
[Bug] Pausing During Warmup Countdown To Knife


PhlexPlexico:
### Expected behavior
When you're calling a timeout during the warmup countdown after both teams `!ready`, it should either not subtract total timeouts, or queue a timeout going into the next state.

### Actual behavior
As soon as the warmup countdown starts (and gamestate changes right before the countdown), if any team calls a `!pause`, it will subtract from their total pauses they have for the half/game, but will not pause the game. 

### Steps to reproduce

Please note that "latest" is NOT a version! See https://github.com/splewis/get5/wiki/Debugging-help#finding-get5-version if you need help finding these. Adding the output file produced by ``get5_debuginfo`` is extremely helpful.

- Plugin version: 0.7.2-dev-358a67a
- Sourcemod version: 1.10
- Steps to reproduce (please be specific):
    - Create a match, get both teams to ready up. Have a predetermined set time for pause, as well as a total amount of pauses.
    - During the warmup countdown once everyone is ready, have a team type `!pause`.
    - Their total pauses will be subtracted by one, but no timeout will occur in the next gamestate.

nickdnk:
@PhlexPlexico Could we not just disable pause during countdown? You can just !pause in the freezetime right after.

PhlexPlexico:
Yup! That is definitely a viable fix for this. Another check that could fix it is to adjust the function here - 

https://github.com/splewis/get5/blob/8ffecad946c8206494fdd3ca856d37d1394f1757/scripting/get5/pausing.sp#L1-L3

But just disabling pausing during countdown from warmup works just as well IMO, I forgot I had reported this haha

PhlexPlexico:
So I was looking at previous tickets here to see if there's anything that can be fixed easily - and ran into this https://github.com/splewis/get5/issues/274

So I guess this is a duplicate issue, and it appears pausing during knife round is intended. But we need to do an additional check I think to see if  maybe something as simple as checking if we are currently in Get5State_GoingLive and making sure we don't fire off the pause events to avoid this bug.

nickdnk:
I think we should just disable pausing during this stage and people can just do it in the freezetime on round one. There's no reason to add complexity to the pausing system when this can be solved by just telling people to wait for the game to start.

PR #784, opened 2022-07-21
Clean up pause time counter logic

nickdnk:
1. Only use one timer for tracking pauses.
2. Ensure fixed pause configuration takes precedence over combined max pause time.
3. Remove ability to call `mp_pause_match` or `mp_unpause_match` directly as those mess with the pause logic
4. Renamed a lot of `g_` variables to be more consistent and removed some that are no longer needed.
5. Renamed `g_MaxTechPauses` to have `g_MaxTechPausesCvar` like the rest of cvars
6. Fixed missing Danish translations
7. Use [PrintHintText()](https://sm.alliedmods.net/new-api/halflife/PrintHintText) instead of in-game pause counters and timers. This makes the code a lot easier to manage and means all pauses are consistently presented the same way. It also allows for unpausing fixed-length pauses, which was otherwise impossible, even for admins.
8. Add `pause_type` to the unpause event object
9. Don't allow calling `!unpause` if a pause was called during a round; you must now wait until freezetime and the pause count will be correctly consumed. The opposing team, however, can call `!unpause` immediately if they just want to unpause as soon as the pausing team does.
10. Don't allow spectators or team_none to ever pause or unpause anything
11. Removed the "each half" suffix from translations related to pauses resetting. It's not consistently applicable across all languages, and I would suggest we replace it with an additional message which contains a universally translate-able string such as "Note that pauses reset on halftime."
12. Make sure to unpause when loading a match config.
13. Don't allow stop command use while admin has paused the game.
14. Fix https://github.com/splewis/get5/issues/647


nickdnk:
Maybe we could also set
https://github.com/saul/demofile/blob/5327326227966b53fbeb08288622f7a8f0329d64/src/sendtabletypes.ts#L4198
when calling a tech pause?

nickdnk:
I also noticed that we currently seem to allow spectator team to both pause and unpause, which could create some problems. I think we need to filter out both team_none and team_spec from both commands entirely. Unless spectators are meant to be able to pause.

PhlexPlexico:
Okay was testing the pauses and first thing broke lol

## Test Case One
Relevant pause vars:
```
get5_max_pause_time "300"
get5_max_pauses "0"
get5_fixed_pause_time "0"
```


This should make all pauses infinite and require intervention on both users to resume, regardless of max pause time I believe? Right now, it allows for 1 timeout of 60 seconds on each side, and trying to call !pause again just simply resets the countdown to round start timer to 15.

Calling a tech pause does the same thing, and the TechPause timer goes up when in freeze time, but stops during round, the match is not paused.

Attached is the server log, will continue with attempting fixed timer max pauses. I will attach this to the issue once I'm done testing as well
[serverlog test case one.log](https://github.com/splewis/get5/files/9164693/serverlog.test.case.one.log)

Also, admins force pausing the match have it failed to pause even with using sm_pause:
```
sm_pause
L 07/21/2022 - 22:16:21: [get5] Calling Get5_OnMatchPaused()
L 07/21/2022 - 22:16:21: [G5WS] Our URL is: https://phlex.avidgamers.me/api/match/%s/pause
L 07/21/2022 - 22:16:21: [G5WS] Trying to create request to url https://phlex.avidgamers.me/api/match//pause
L 07/21/2022 - 22:16:21: [get5] Calling Get5_OnEvent(data={
    "matchid": "3833",
    "team": null,
    "event": "game_paused",
    "map_number": 0,
    "pause_type": "admin"
})
[♣Get5☺] An admin force paused the match.
Preventing call to mp_pause_match. Administrators should use sm_pause/sm_unpause.
```

### Test Case Two
45 second `!pause` with no limits on pausing is working as expected. However, tech pauses remaining broken in this test, as well as admin/rcon servers calling `sm_pause`. Attached is log file.

```
get5_max_pause_time "0"
get5_max_pauses "0"
get5_fixed_pause_time "45"
```
[serverlog_fixed_pause_time_45.log](https://github.com/splewis/get5/files/9164700/serverlog_fixed_pause_time_45.log)

### Test Case Three
get5_max_pause_time "0"
get5_max_pauses "4"
get5_fixed_pause_time "45"

The first timeout is incorrect as it's showing 0/4 timeouts, second timeout shows 2/4. Admins are unable to pause and tech pauses do not happen but the counter is shown to be counting upwards in the LogDebugs until match start, or until someone else calls a `!pause`.



Attached below is the log file.
[serverlog_fixed_pause_time_45_max_4.log](https://github.com/splewis/get5/files/9164711/serverlog_fixed_pause_time_45_max_4.log)

I think hooking into the mp_pause/unpause may not be the play here, due to the fact ServerCommand will always call the hook which immediately prevents it from working as intended. I believe all of the following test cases should work as intended: 

```
get5_max_pause_time "300"
get5_max_pauses "0"
get5_fixed_pause_time "0"
```
- This should allow a timeout to last a total of 300 seconds without intervention before the match unpauses, an unlimited number of times. Tech pauses and otherwise should not be included in this as they are separate cvars, that only interact with hte base (old `stock`) pause and unpause calls. I believe this one can _not_ use the built in UI, due to the fact we have the timer running and waiting on chat feedback, plus I don't think you can cancel the call once it's on the UI (I've not tested this thoroughly, though).

```
get5_max_pause_time "0"
get5_max_pauses "0"
get5_fixed_pause_time "45"
```
- This should allow unlimited 45 second fixed pause timers. While similar to the above case, this will not allow users to call `!unpause` and stop the count of the 300 seconds, and uses the built in UI.

```
get5_max_pause_time "0"
get5_max_pauses "0"
get5_fixed_pause_time "0"
```
- This should essentially act as a "tech" timeout so to speak as users both need to confirm to continue, or be unpaused by an admin. I don't think the UI/CS:GO supports this, so this would have to be done with chat commands.

```
get5_max_pause_time "0"
get5_max_pauses "4"
get5_fixed_pause_time "0"
```
- This should allow 4 unlimited time pauses, where both users must call `!unpause`, or be unpaused by an admin.

```
get5_max_pause_time "300"
get5_max_pauses "4"
get5_fixed_pause_time "0"
```
- Last test case, is that a user should have a pause time of 300, but be able to resume whenever (since UI doesn't support this?), up to 4 times total.


The issue right now with this PR as well, is that the underlying `stock bool Pause(...` and `stock bool Unpause(...` are both modified, which inherently changes how the tech pausing works in game as well. I think we should maybe take a step back and try altering only that pause timer for tech pauses and get that working, then maybe continue to experiment with what CS:GO has to offer for tech pauses as well.

I hope this test case helps, and if you need me to test more I can rig up these cases again and keep testing for intended behaviour :)

Thanks!

PhlexPlexico (approved):
LGTM! After discord talks and testing, we even squashed a few bugs in relation to the UI timeout not updating properly. Looks a lot cleaner and everything is working as intended.

nickdnk:
So I'm fairly certain everthing works as we want it to now. Please give this a run, @PhlexPlexico. I've tested it pretty extensively myself.

Remember to copy in translations; I added a bunch more.

scripting/get5/matchconfig.sp
```diff
@@ -27,12 +27,11 @@ stock bool LoadMatchConfig(const char[] config, bool restoreBackup = false) {
...
-      g_TeamPauseTimeUsed[team] = 0;
-      g_TeamPausesUsed[team] = 0;
-      g_TeamTechPausesUsed[team] = 0;
+      g_TacticalPauseTimeUsed[team] = 0;
+      g_TacticalPausesUsed[team] = 0;
+      g_TechnicalPausesUsed[team] = 0;
     }
-    g_TechPausedTimeOverride[team] = 0;
-    g_TeamGivenTechPauseCommand[team] = false;
+    g_TechnicalPauseTimeUsed[team] = 0;
```
nickdnk, line 34: Why is this outside here? Is this intentional?

PhlexPlexico: Yes, `!stop` I believe can be called during a tech timeout, so if we do a round restore, we still wish to zero out the time because we will enter a different pause state. I believe prior to refactoring there would be issues surrounding timers on restore.

PhlexPlexico:
Overall this still looks good to me. I'll test this tomorrow.

In regards to the inconsistencies with tech pause time tracking and max tactical pauses, this is intentional. I believe the tracked timing is more of an independent Get5 features, where technical pauses already had precedence in the way it's implemented (right now, it's identical to ESEA), and was requested to be implemented like this by users. So I believe we should keep it as is :)

scripting/get5/matchconfig.sp
nickdnk: The point here is that `g_TechnicalPauseTimeUsed[team] = 0;` is always reset when a tech pause ends. It doesn't stack across multiple tech pauses, so why is it explicitly being skipped here?

https://github.com/splewis/get5/pull/784/files#diff-80fe068e671d54322735140aeae59d9c10ffea3db4bb6004365876b596355f3fR28-R35

nickdnk: Ah sorry I've inverted it in my understanding. It **always** gets reset here also. Never mind. It's good.

nickdnk:
Okay, so we can get rid of the array-logic for technical pause time and just use a single counter like we do with the fixed pause timer. If it's not supposed to track time used per team, we don't need another variable for it.

PhlexPlexico (approved):
LGTM 👍

scripting/get5/pausing.sp
```diff
@@ -1,8 +1,17 @@
 public bool Pauseable() {
-  return g_GameState >= Get5State_KnifeRound && g_PausingEnabledCvar.BoolValue;
+  return g_GameState == Get5State_Live && g_PausingEnabledCvar.BoolValue;
```
splewis, line 2: > Game can now only be paused when live. If you want a pause during knife, you must wait until live freezetime. I'm not 100% sure we want this, but it seems like calling a pause right after you readied your team is stupid. Admin can always pause.

Things can change in the 20 seconds between the match officially begins and the knife freeze time. I would rather this change not be made.

scripting/get5/pausing.sp
```diff
@@ -1,8 +1,17 @@
...
 }
 
-public bool PauseGame(Get5Team team, Get5PauseType type, int pausesLeft) {
+stock void PauseGame(Get5Team team, Get5PauseType type) {
```
splewis, line 5: Doesn't really matter, but it would make more sense to remain 'public'.

translations/da/get5.phrases.txt
```diff
@@ -6,6 +6,7 @@
     }
     "WaitingForCastersReadyInfoMessage"
     {
+        "#format"       "{1:s}"
```
splewis, line 9: Why are these being added? They should typically only be on the english file: https://wiki.alliedmods.net/Translations_(SourceMod_Scripting)#Distributing_Language_Files

nickdnk: Didn't know. I'll revert that.

scripting/get5/pausing.sp
nickdnk: I think it's also fine to let this go back to >=knife. I originally made this change to reduce logical complexity when refactoring, but it's been cleaned up so much that I don't see a problem with this. I'll just have to test it once.

PhlexPlexico: I think one other test that should be done is to check the warmup countdown right after the knife decision, as that reduces the amount of fixed pauses a team has per half, as reported here, without pausing the game - https://github.com/splewis/get5/issues/647 so maybe test this as well, just in case this still happens then? I think with this refactoring, that may no longer be an issue?

scripting/get5/pausing.sp
nickdnk: Was changed because it had default values which are unsupported when using public. It doesn't anymore though, after I removed in-game-state pauses, so I can change it back. I still don't understand the difference entirely. It's not a callback and it's not being used outside the mod.

Only obvious advantage I see to public here is that we get the "not used" warning from the compiler if we someway don't use it.

I'll change it back.

scripting/get5/pausing.sp
nickdnk: The problem is that the game is restarted when the knife round ends, which I think unpauses? Will have to test.

It might actually be a bit of a problem to work around if that is in fact the case. @splewis is it a big deal to just let people pause in the freeze time after going live? They're going to end up paused at the same stage anyway, they just can't do it **while** still knifing for sides.

nickdnk: Oh, the problem is of course if you want to pause in the **knife** freezetime? Because yea, we wouldn't allow that currently.

nickdnk: Might be solvable by changing the if to `if (g_GameState >= Knife && g_GameState != WaitingForKnifeDecision)`.

nickdnk: @splewis Fixed. You can pause during live, knife and waiting for knife decision and going live. I tested for https://github.com/splewis/get5/issues/647 and it works in all cases I can think of.

PR #786, opened 2022-07-24
Add function to check for latest version of get5

nickdnk:
This allows get5 to check if its version is the same one as the one on the master branch.

~~We need to do some regex stuff after the CI stuff has been merged to make this work. Currently it matches against the `PLUGIN_VERSION` 1:1 on master which would also be incorrect as it includes `-dev`.~~ Edit: I fixed this.

I tested this and it works. It also correctly doesn't load if SteamWorks is not installed.

What I don't know is if I need to `delete` the `request` handle in the callback function.

Output example would be, with `0.9.0-dev` in the `version.sp` file on master and `0.8.0` in the official version used:

`[get5.smx] A newer version of Get5 is available. You are running 0.8.0 while the latest version is 0.9.0.`

### DIFF
--- a/scripting/get5/version.sp
+++ b/scripting/get5/version.sp
@@ -2,3 +2,8 @@
 #if !defined PLUGIN_VERSION
 #define PLUGIN_VERSION "0.9.0-dev"
 #endif
+
+// This MUST be the latest version in x.y.z semver format followed by -dev.
+// If this is not consistently applied, the update-checker might malfunction.
+// In official releases, the CI flow will remove the -dev suffix when compiling the plugin.
+// In development pre-releases, dev is replaced with the first 6 characters of the commit hash.
--- a/documentation/docs/commands.md
+++ b/documentation/docs/commands.md
@@ -83,7 +83,7 @@ directory.
 
 ####`get5_loadmatch_url <url>` {: #get5_loadmatch_url }
 :   Loads a remote (JSON-formatted) match config by sending an HTTP(S) GET to the given URL. This requires the
-[Steamworks](../installation/#installing-steamworks-optional) extension. When specifying a URL with http:// or
+[SteamWorks](../installation/#steamworks) extension. When specifying a URL with http:// or
 https:// in front, you have to put it in quotation (`""`) marks.
 
 ####`get5_endmatch`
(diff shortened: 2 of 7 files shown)


nickdnk:
I did consider if we should have a `g_NewerVersionAvailable` global variable that stores a newer version and then print `A newer version of Get5 is available` in chat when the game goes live, similarly to where it says `Powered by Get5`. At least it would be seen way more often there.

nickdnk:
Screenshot from 2022-07-24 22-24-28

I added to the game, it would look something like this and be localizable. It tested it and it works fine.

splewis:
I feel like the in-game chat message is unnecessary, and more likely to annoy players rather than notify the actual server admin in many cases -- especially if it's not easily disabled. How about just logging it?

nickdnk:
> I feel like the in-game chat message is unnecessary, and more likely to annoy players rather than notify the actual server admin in many cases -- especially if it's not easily disabled. How about just logging it?

It's not likely to _annoy_ anyone since it does not modify the game in any way or prevent anyone from doing what they want to do. What it does, however, is nudge people to update the plugin. They can still choose not to, but it's in our interest (and theirs) that they don't run outdated software. It already logs to console, but I think it will be very often missed here vs. showing it in the game.

From the perspective of someone who does not ever want to update the plugin, yes, this might be "annoying", but why would you not want to do that? I can add a convar for disabling the check as a middle-ground, but I think this is one of those times where asking what the user "wants" is not the right call.

nickdnk:
Another middle ground could be making the message shorter. We could drop the version codes from it and just have it say `A newer version of Get5 is available. Visit splewis.github.io/get5 to update.`

nickdnk:
Added Cvar so it can be disabled if someone would want to do that.

nickdnk:
This is runnning an official build that's outdated:
Screenshot from 2022-07-26 01-54-01

This is running a pre-release off development or if compiled from source:
Screenshot from 2022-07-26 01-52-32

scripting/get5/version.sp
```diff
@@ -1,4 +1,9 @@
 #tryinclude "manual_version.sp"
 #if !defined PLUGIN_VERSION
-#define PLUGIN_VERSION "0.9.0-dev"
+#define PLUGIN_VERSION "0.8.1"
```
nickdnk, line 3: This is obviously not supposed to change. Was testing locally. Will fix.

splewis:
I like the shorter message + cvar. Feel free to go ahead and merge this.

PR #789, opened 2022-07-26
Better bot support + rehook SDKHooks on plugin reload

nickdnk:
This allows bots to be included in the damage report. It also works correctly if a player controls a bot.

I fixed the missing SDKHook on plugin reload, which was the cause of the issues I had with this hook not working.

Also, prevent the spam of auth failure when bots deal damage etc.

Edit: Accidentally had the wrong base. it's good now (on develop). I tested this and it works fine.

### DIFF
--- a/scripting/get5/stats.sp
+++ b/scripting/get5/stats.sp
@@ -20,12 +20,11 @@ public void Stats_PluginStart() {
 
 public Action HandlePlayerDamage(int victim, int &attacker, int &inflictor, float &damage,
                           int &damagetype) {
-  LogDebug("HandlePlayerDamage(victim=%d, attacker=%d, inflictor=%d, damage=%f, damageType=%d)",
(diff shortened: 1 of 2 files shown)


scripting/get5/stats.sp
```diff
@@ -1137,47 +1139,52 @@ static void PrintDamageInfo(int client) {
     return;
 
   char message[256];
+  int msgSize = sizeof(message);
```
PhlexPlexico, line 1142: I know that the whole `sizeof(message)` gets to be repeated a lot, but this is the first time it appears in any of the codebase. Either we refactor everywhere (could be done in separate commit), or we change styles to have this for every `sizeof` call that's used more than once in a function call?

PhlexPlexico (approved):
Aside from the one comment, that looks good and I'm happy with this moving forward as well!

Issue #344, opened 2019-02-04
Backup file format/directory cvar


FlowingSPDG:
Is there's cvar to configure get5's backup file format/directory,it woulbe awesome fore me :)
Currently I'm running SRCDS on AWS,but sharing backup file is not easy for now.
So,I want to save backup files into NFS(EFS for AWS) shared storage.
(e.g. creating "backup" symlink folder)
thanks :) and sorry for poor English skill.

splewis:
If someone wants to implement this, I added some notes at https://github.com/splewis/get5/issues/335#issuecomment-456200592 on the special concerns for it.

nickdnk:
Will be fixed in https://github.com/splewis/get5/pull/793.

PR #793, opened 2022-07-28
Allow for backup prefix

nickdnk:
Inspired by https://github.com/splewis/get5/pull/650, this should allow you to put backups into subfolders on your server. Not thoroughly tested yet.

Not meant to be included in 0.9. Just a proposal here.

### DIFF
--- a/documentation/docs/backup.md
+++ b/documentation/docs/backup.md
@@ -22,7 +22,7 @@ respectively. A special backup called `get5_backup_match%s_map%d_prelive.cfg` is
 When in a match, you can call [`get5_listbackups`](../commands/#get5_listbackups) to view all backups for the current
 match. Note that all rounds and map numbers start at 0.
 
-They print in the format `filename date team1 team2 map team1_score team2_score`.
+They print in the format `filepath date time team1 team2 map team1_score team2_score`.
 
 ```
 > get5_listbackups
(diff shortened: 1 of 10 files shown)


PhlexPlexico:
I think #344 should be linked for this as well.

nickdnk:
Yes, I did also add code to make the list function work with prefixes.

nickdnk:
Fixes https://github.com/splewis/get5/issues/344.

nickdnk:
This should be ready for some testing if you feel like it @PhlexPlexico.

PhlexPlexico:
I'll get my machines up tomorrow here and test this as well! Sorry it's taken so long

nickdnk:
So the current commit allows directories in the prefix, but only if they already exist. I'm working on some code that recursively creates directories, but I'm not quite done yet.

nickdnk:
As of this comment, this seems to work pretty well.

It supports creating nested subfolders, including with the {MATCHID} variable, i.e. `backups/{MATCHID}/`, and both the `get5_loadbackup` and `get5_listbackups` functions work fine with this.

I've renamed it to `get5_backup_path`, as you **must** include a forward slash in the variable, so it cannot be used just as a file prefix. It must be a subfolder structure.

Please give it a go.

PhlexPlexico (approved):
Tested on backups and restores, and it seems to be working for me.

nickdnk:
I just adjusted the code that handles automatic deletion of backups so it still works.

Also stressed in the docs the importance of a trailing slash in the backup path.

PR #798, opened 2022-07-29
Simplify suicide logic

nickdnk:
Suicide logic was overly complex. In the end all we needed to do was filter out the bomb.


nickdnk:
Tested working.

PR #797, opened 2022-07-29
Added Hungarian translation

enerbewow:

PR #803, opened 2022-07-30
0.9 Release

nickdnk:
@splewis Final review before 0.9 release.

I will bring in the Hungarian translation (https://github.com/splewis/get5/pull/797) once tested, but that's only a translation file edit. Nothing else will change unless we find a problem that needs fixing.

Key points:

1. Refactored the pause system. It has been very thoroughly tested and works well.
2. Fixed a bunch of problems related to coaching, including the game-breaking bug that would [put coaches in 3rd person](https://github.com/splewis/get5/pull/803/commits/f7b5c0d44da93472035ec539976d78ca9a3accdf). Valve couldn't fix this it seems, but @PhlexPlexico and I figured it out.
3. A lot better support for and handling of bots, including putting them in the damage report.
4. Fixed the problem with scrims and manual matches messing up the MySQL extension as it expects only integer Match IDs.
5. Fixed some problems related to suicide and dying from the bomb. Also thoroughly tested.
6. Streamlined CI flow to allow for pre-releases (development branch) and releases (master branch).
7. Fixed a bunch of translation-related stuff.
8. Fixed a bug that would cause damage attribution (damage report + grenade victim forwards) to not work if the plug-in was reloaded while users were on the server.
9. Brought in the update-notice functionality.

This is way better than the current 0.8 master release (which also has no official release), so I suggest we get this released ASAP.


splewis:
Looks good to me; let's merge it in.

PhlexPlexico (approved):
LGTM 🎉

PR #805, opened 2022-07-31
Add hungarian translation

nickdnk:
Messed up the 0.9 merge, so this is a replacement PR for https://github.com/splewis/get5/pull/797

@enerbewow: Download test version with Hungarian translation here: https://github.com/splewis/get5/actions/runs/2767755954

### DIFF
--- a/translations/hu/get5.phrases.txt
+++ b/translations/hu/get5.phrases.txt
@@ -1,0 +1,387 @@
+"Phrases"
+{
+    "ReadyToVetoInfoMessage"
+    {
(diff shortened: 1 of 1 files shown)


enerbewow:
I tested it and works smooth. One little thing.

When I did something unusual (mp_maxround 6 when I was in the 4th round) and when I won the final round, the end game announcement isn't worked. TeamWonMatchInfoMessage
[Get5] Won the match without the team name

enerbewow:
In my case:
[Get5] megnyerte a mérkőzést.

enerbewow:
I did a match with current players on server

nickdnk:
Alright. That's a separate issue though, as your translation file is correct. We will have to file an issue for that specifically.

PR #806, opened 2022-07-31
Reorder parameters of get5_creatematch to match get5_scrim

nickdnk:
Because an empty match is default, and required if using the MySQL extension, the parameter order of `get5_creatematch` should match that of `get5_scrim` to put `matchid` last.


nickdnk:
Accidentally committed these changes to develop with the documentation update for it. So it's merged.

PR #810, opened 2022-07-31
Add in checks for SVG logos

### LINKED ISSUE #807: [BUG] SVGs not added to download table
If making a feature request, you should delete all the pre-filled text here.
If reporting a bug, fill in the following sections:

### Expected behavior
If a user adds new logos to a team when setting up a match (provided they have a FastDL server), the SVG logos should be added to the download table.

### Actual behavior
SVGs are not added, as the code only allows for PNGs.

### Steps to reproduce
- Set up a server that contains custom SVG logos, and ensure they exist on the FastDL server. 
- Create a match config with these logos.
- Load Match Config
No info will be in the console if those team logos have been added to the download table or not. This is due to this call not containing checks for SVG, only PNGs. 

https://github.com/splewis/get5/blob/112b52b99989835891438335c982827eab25f4b3/scripting/get5/matchconfig.sp#L1233
Please note that "latest" is **NOT** a version! In order to get the version, please use `get5_version` in the server console, or `get5_debuginfo` and attach the file located at `addons/sourcemod/logs/get5_debuginfo.txt`.

- Plugin version: 0.9.0
- Sourcemod version: 1.10

All that should be done is a check for `FileExists` either on the PNG first, or SVG (different path, as well) first, then default to the other. If neither exist, log an error and explain the file isn't found.

PhlexPlexico:
Currently the `AddTeamLogoToDownloadTable` does not work as intended since the Panorama update. Currently, it only checks for PNG files located in one specific directory (which was used for ScaleForm UI), and will only show in a few places. The new SVG format has more usage in game, so we need to check if we're using an SVG instead.

This closes #807

### DIFF
--- a/scripting/get5/matchconfig.sp
+++ b/scripting/get5/matchconfig.sp
@@ -1230,10 +1230,20 @@ static void AddTeamLogoToDownloadTable(const char[] logoName) {
     return;
 
   char logoPath[PLATFORM_MAX_PATH + 1];
-  Format(logoPath, sizeof(logoPath), "resource/flash/econ/tournaments/teams/%s.png", logoName);
-
-  LogDebug("Adding file %s to download table", logoName);
-  AddFileToDownloadsTable(logoPath);
+  Format(logoPath, sizeof(logoPath), "materials/panorama/images/tournaments/teams/%s.svg", logoName);
+  if (FileExists(logoPath)) {
+    LogDebug("Adding file %s to download table", logoName);
+    AddFileToDownloadsTable(logoPath);
+  } else {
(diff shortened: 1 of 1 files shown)


nickdnk (approved):

nickdnk:
And this works with both custom and already-included team files?

PhlexPlexico:
Already included should have no effect, since they won't be added to the download table (existing in game by default). This still doesn't fix the issue with custom logos needing a game restarted twice, that fix will be coming soon™

nickdnk:
Alright.

PR #812, opened 2022-07-31
Live Coach Fix

### LINKED ISSUE #811: [BUG] Coaches Disconnecting and Reconnecting During Live
### Expected behavior
When a coach disconnects and reconnects in a live match (i.e. no freeze time), they should be dumped back onto the coaching slot.

### Actual behavior
A user is set at the teamjoin menu, and they are able to press escape to get to the default spectator camera view, introducing the dreaded camera bug.

### Steps to reproduce
- Create match
- Let user join a coach position
- Begin match, while live have the coach disconnect and reconnect
- Greeted with `You can only change coaching positions during warmup` text.
- User can press escape to see default camera view.

- Plugin version: 0.9.0
- Sourcemod version: 1.11

This is an easy fix at least, because we can also just check if the gamestate is currently live and we're not paused, to force them back into the coach spot with the right spectator mode set.

PhlexPlexico:
Swap coaching logic to use in-game coaching feature only in warmup (as provided by valve), since that is what the error message provides.

This update now moves coaches in all cases, since when a player joins and if they exist in the coach array, we will move them to the coach slot no matter the case.

Backups are also working as intended, where the coach is moved back into the slot with no changes to money.

Attempting to join during a live scrim, a user must call .coach if get5_check_auths is 0, but the player will still reliably be placed on the coach slot, with the proper camera.

Putting this one in draft as I would like to test a few more things before having it marked as ready.

Closes #811

### DIFF
--- a/scripting/get5/teamlogic.sp
+++ b/scripting/get5/teamlogic.sp
@@ -162,18 +162,19 @@ public void MoveClientToCoach(int client) {
   // If we're in warmup we use the in-game
   // coaching command. Otherwise we manually move them to spec
   // and set the coaching target.
-  // If in freeze time, we have to manually move as well.
-  if (!InWarmup() && InFreezeTime()) {
+  // If in freeze time, or any other state like a live round,
+  // we have to manually move as well.
+  if (InWarmup()) {
+    LogDebug("Moving %L indirectly to coach slot via coach cmd", client);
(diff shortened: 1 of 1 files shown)


PhlexPlexico:
Moving into #815 since both changes kind of go hand-in-hand for team logic changes.

PR #815, opened 2022-08-01
Simplify CheckClientTeam and Update Coach Placement

PhlexPlexico:
Users were not actually being moved when UpdateCoachTarget was called, so we should use our helper function CheckIfClientCoachingAndMoveToCoach which does all the setting and placement instead. If they aren't coaching then switch player team if they are not on the right team.

Swap coaching logic to use in-game coaching feature only in warmup (as provided by valve), since that is what the error message provides.

This update now moves coaches in all cases, since when a player joins and if they exist in the coach array, we will move them to the coach slot no matter the case.

Backups are also working as intended, where the coach is moved back into the slot with no changes to money.

Attempting to join during a live scrim, a user must call .coach if get5_check_auths is 0, but the player will still reliably be placed on the coach slot, with the proper camera.

Putting this one in draft as I would like to test a few more things before having it marked as ready.

Closes https://github.com/splewis/get5/issues/811

### DIFF
--- a/scripting/get5/teamlogic.sp
+++ b/scripting/get5/teamlogic.sp
@@ -18,21 +18,10 @@ public Action Timer_PlacePlayerOnJoin(Handle timer, int userId) {
 
 public void CheckClientTeam(int client) {
   Get5Team correctTeam = GetClientMatchTeam(client);
-  char auth[AUTH_LENGTH];
   int csTeam = Get5TeamToCSTeam(correctTeam);
   int currentTeam = GetClientTeam(client);
 
-  if (csTeam != currentTeam) {
-    if (IsClientCoaching(client)) {
-      UpdateCoachTarget(client, csTeam);
-    } else if (GetAuth(client, auth, sizeof(auth))) {
-      char steam64[AUTH_LENGTH];
-      ConvertAuthToSteam64(auth, steam64);
-      if (IsAuthOnTeamCoach(steam64, correctTeam)) {
-        UpdateCoachTarget(client, csTeam);
-      }
-    }
-
+  if (!CheckIfClientCoachingAndMoveToCoach(client, correctTeam) && csTeam != currentTeam) {
     SwitchPlayerTeam(client, csTeam);
   }
 }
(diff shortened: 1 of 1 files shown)


scripting/get5/teamlogic.sp
```diff
@@ -18,22 +18,16 @@ public Action Timer_PlacePlayerOnJoin(Handle timer, int userId) {
...
   if (csTeam != currentTeam) {
-    if (IsClientCoaching(client)) {
-      UpdateCoachTarget(client, csTeam);
-    } else if (GetAuth(client, auth, sizeof(auth))) {
-      char steam64[AUTH_LENGTH];
-      ConvertAuthToSteam64(auth, steam64);
-      if (IsAuthOnTeamCoach(steam64, correctTeam)) {
-        UpdateCoachTarget(client, csTeam);
-      }
+    if (!CheckIfClientCoachingAndMoveToCoach(client, correctTeam)) {
```
nickdnk, line 25: Since you run `CheckIfClientCoachingAndMoveToCoach` in any case, this whole block can be changed to:

```c
if (!CheckIfClientCoachingAndMoveToCoach(client, correctTeam)) {
  if (csTeam != currentTeam) {
    SwitchPlayerTeam(client, csTeam);
  }
}
```

scripting/get5/teamlogic.sp
```diff
@@ -169,17 +155,17 @@ public void MoveClientToCoach(int client) {
   // coaching command. Otherwise we manually move them to spec
   // and set the coaching target.
   // If in freeze time, we have to manually move as well.
-  if (!InWarmup() && InFreezeTime()) {
+  if (InWarmup()) {
+    LogDebug("Moving %L indirectly to coach slot via coach cmd", client);
```
nickdnk, line 159: Why is this "indirectly"? Wouldn't it just be "Moved to coach slot with coach command"?

PhlexPlexico: I think this is because we are making the client indirectly call `coach`, that's why. We can change it, but this was just what was here before I modified anything.

Issue #535, opened 2020-06-02
Get5 API not downloading missing logos


tuudik:
# Expected behavior
When loading in match config over API, missing logos (SVG or PNG, depending of config), will be downloaded.

### Actual behavior
When loading in match config over API, missing logos (SVG or PNG, depending of config), are not downloaded.

CheckForLogo function will just not work, since the passed logo value is empty.

### Steps to reproduce

To test it I added LogDebug lines like following to get5_apistats.sp:
```
char logo1[32];
  char logo2[32];
  GetConVarStringSafe("mp_teamlogo_1", logo1, sizeof(logo1));
  GetConVarStringSafe("mp_teamlogo_2", logo2, sizeof(logo2));
  LogDebug("Checking for logo: %s", logo1);
  LogDebug("Checking for logo: %s", logo2);
  CheckForLogo(logo1);
  CheckForLogo(logo2);
}

public void CheckForLogo(const char[] logo) {
  if (StrEqual(logo, "")) {
    return;
  }
  LogDebug("Checking for logo: %s", logo);
  char logoPath[PLATFORM_MAX_PATH + 1];
```

I turned on the get5_debug 32 and checked get5_debug log:
```
L 06/02/2020 - 17:58:55: [get5] Calling Get5_OnSeriesInit
L 06/02/2020 - 17:58:55: [get5_api] mp_teamlogo_1: 
L 06/02/2020 - 17:58:55: [get5_api] mp_teamlogo_2: 
L 06/02/2020 - 17:58:55: [get5] Adding file 3dm to download table
L 06/02/2020 - 17:58:55: [get5] Adding file astr to download table
```

Since the mp_teamlogo_1 and mp_teamlogo_2 were empty, the CheckForLogo function didnt work.

- Plugin version: 0.7.2-dev 8890db8
- Sourcemod version: 1.10

tuudik:
@MoritzLoewenstein This is what and how I've discovered :)

MoritzLoewenstein:
Thanks, will look into it.

tuudik:
Well yeah, the AddTeamLogoToDownloadTable function matchconfig.sp works :)

tuudik:
Added some additional debug lines to get5, can see that `mp_teamlogo_*` ConVar's are set after not before the api is asking for it:
```L 06/03/2020 - 09:12:16: [get5] Calling Get5_OnEvent(event name = series_start)
[...]

PR #816, opened 2022-08-02
Move Match Cvars above series init 

### LINKED ISSUE #161: Custom logo
Hello,

I made earlier an issue on get5-web and you told me I could put the logo in csgo/resource/flash/econ/tournaments/teams/
So I did, I put my logo ar.png in it.

Then I made my team_argon file:
https://hastebin.com/zidanekewu.pl

I tried: 
- ar
- /home/steam/5v5/csgo/resource/flash/econ/tournaments/teams/ar.png
- ar.png

Nothing works, it just shows empty logo.

I searched on the wiki but couldn't find how to do.

Thanks in advance,
sach.

PhlexPlexico:
This should allow certain vars to be set before the event is called, where users can then take advantage of having the logos set properly or have them be downloaded to the server.

Just looking through the code, I think this may be inconsequential for match loading, but I think more extensive testing needs to be done.

Closes #161 #535

### DIFF
--- a/scripting/get5/matchconfig.sp
+++ b/scripting/get5/matchconfig.sp
@@ -116,6 +116,13 @@ stock bool LoadMatchConfig(const char[] config, bool restoreBackup = false) {
     ChangeState(Get5State_PreVeto);
   }
 
+
+  // We need to ensure our match team CVARs are set
+  // before calling the event so we can grab values
+  // that are set in the OnSeriesInit event.
+  // Add to download table after setting.
+  SetMatchTeamCvars();
+
   if (!restoreBackup) {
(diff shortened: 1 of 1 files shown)


PhlexPlexico:
As discussed in the Discord. It appears there is a limitation of logos in CS:GO. For anyone looking back and wondering why "logos aren't working", it is because you must _restart your client_ entirely in order for logos to be picked up. Server owners who setup FastDL servers and have logos in their matches will have everything working as intended, (being added to download table and all), but users will not be able to see these logos until they join the server (to download logos), then exit and restart their client. Dumb limitation, but it seems to work when these steps are done.

Issue #817, opened 2022-08-02
Backup should store draw info


nickdnk:
I was working on the [backup path thing](https://github.com/splewis/get5/pull/793), when I came across this. It would appear that game series scores are saved to backups and also restored, but that this does not include draws.

Written to backup here:
https://github.com/splewis/get5/blob/b8d6aff3a3796ac594ad5288d9265aea102b4f5d/scripting/get5/backups.sp#L164-L165

Read from backup here:
https://github.com/splewis/get5/blob/b8d6aff3a3796ac594ad5288d9265aea102b4f5d/scripting/get5/backups.sp#L246-L247

All three reset to 0 here, when loading a match config during restore:

https://github.com/splewis/get5/blob/b8d6aff3a3796ac594ad5288d9265aea102b4f5d/scripting/get5/matchconfig.sp#L22-L23

And used here to calculate `Get5_GetMapNumber`:

https://github.com/splewis/get5/blob/b8d6aff3a3796ac594ad5288d9265aea102b4f5d/scripting/get5/natives.sp#L203-L206

... so when a game is restored and a map was a draw, it's going to be wrong. Unless I'm missing something here.

PhlexPlexico:
When loading a match, yes, but these values are brought in through `LoadTeamData`, as seen here.

https://github.com/splewis/get5/blob/b8d6aff3a3796ac594ad5288d9265aea102b4f5d/scripting/get5/matchconfig.sp#L655

Even though they get zeroed out, they are reinstantiated.

How exactly is it calculated on match draw when the map ends? Would it be set to 1 and 1 for each team? I can see the restore will always get it in the LoadTeamData call that's called from `LoadMatch`.

PhlexPlexico:
Also I think there are no draws in competitive play, there is always a result, no matter the amount of overtime played, right?

nickdnk:
> Also I think there are no draws in competitive play, there is always a result, no matter the amount of overtime played, right?

You can disable overtime and just play 15-15.

I see the team info being set, yes, but draws are attributed as wins to `Get5Team_None`. I don't see how they are loaded with the code you linked?

deLuXe83:
Hi,
I can confirm this bug!

Had it multiple times - matches drawed, but stats are not recorded correct.

See for example matches 53, 55 & 57; all bo2 matches!
But the stats could mixed up. It always happen, when first map ended as draw!

[...]

Issue #728, opened 2022-01-01
New Gamemode: not "best-of" but "fixed maps"


deLuXe83:
Hello everybody and a happy new year,

would it be possible to add a "new gamemode" which is not played in a best-of series but in a fixed number of maps!?
Meaning you can determine the number of maps like it is possible right now with "num_maps", but this number of maps is played in any case and the game will not be aborted if one team has no more chance to win the whole game.
In this mode both functions "veto from mappool" and "fixed maplist" should work.

A cool addition would be if you could add a new mode for team choosing. 
For example with only one knife round in the very first map. Winner of this knife round chooses starting side. In the next map there is no knife round, but swapping/alternating of starting sides (eg. if team1 won the knife round and chose to start as CT, in the next map team1 is starting T. In the following map again as CT, then T, and so on...).

What do you think?

Bests!

Issue #213, opened 2017-12-13
Get5_OnSeriesResult is called much later than it could be


splewis:
It gets called in EndSeries, which is typically called after the GOTV delay expires. It should be called much earlier so stats can be updated more quickly.

burhan-kun:
Yes, thats right. After i adding GO TV i become that error. How i can fix this?

splewis:
The only 'fix' is a code change to the get5 plugin. Feel free to send a pull request.

Issue #190, opened 2017-10-09
Set Match winner


keeradfun:
Feature Request.

Can you add a cvar or a chat command to set the winner in between games?
Like if a team forfeit or gets disqualified in the game admin should have access to set winner.

keeradfun:
Hey, i found a csgo cvar command to set winner forcefully mp_forcewin....  it might come in handy to develop this feature...

PR #821, opened 2022-08-07
Ensure correct MapNumber logic when restoring from backups with draws + add series clinch config parameter + fix OnSeriesResult

nickdnk:
This should fix https://github.com/splewis/get5/issues/817.

I also adjusted `Get5_OnBackupRestore` to have the correct map and round number, indicating the match, map and round being restored **to**.

Tried to reduce use of `Get5_GetMapNumber()` where it is not needed.

Fixed problems with ties not correctly affecting the next map/series end logic.

I put the fix for https://github.com/splewis/get5/issues/822 in here as well.

Added a `g_SeriesCanClinch` option that allows you to disable clinch-logic in a series, similarly to `mp_match_can_clinch`, but for maps.

I've moved a lot of code around, fixing the delay on `OnSeriesResult` and probably fixing a lot of weird issues with GOTV recordings. `mp_match_restart_delay` is now corrected if it is not long enough, but not shortened if longer. This allows admins to set a longer, fixed pause length between maps regardless of the length of the GOTV delay. Essentially this now includes https://github.com/splewis/get5/pull/825, but I had to put that on top of this due to the way map scores are now handled. This also gets rid of `g_ForceWinnerSignal` and `g_ForcedWinner`.

Everything here is tested and it all works as far as I can tell.


scripting/get5/backups.sp
```diff
@@ -292,11 +300,13 @@ public bool RestoreFromBackup(const char[] path) {
     kv.GoBack();
   }
 
+  int roundNumberRestoredTo = -1; // Assume -1 for non-live backup
   char tempValveBackup[PLATFORM_MAX_PATH];
   GetTempFilePath(tempValveBackup, sizeof(tempValveBackup), TEMP_VALVE_BACKUP_PATTERN);
   if (kv.JumpToKey("valve_backup")) {
     g_SavedValveBackup = true;
     kv.ExportToFile(tempValveBackup);
+    roundNumberRestoredTo = kv.GetNum("round", 0) - 1; // Round number is not zero-indexed inside valve backup
```
nickdnk, line 309: Need to verify that this is correct; that the round number actually matches the one we restored to.

scripting/get5/matchconfig.sp
```diff
@@ -104,19 +104,21 @@ stock bool LoadMatchConfig(const char[] config, bool restoreBackup = false) {
...
     ChangeState(Get5State_Warmup);
-
-    char currentMap[PLATFORM_MAX_PATH];
-    GetCurrentMap(currentMap, sizeof(currentMap));
-    if (!StrEqual(mapName, currentMap) && !restoreBackup) {
-      ChangeMap(mapName);
+    if (!restoreBackup) {
+      // When restoring from backup, changelevel is called after loading the match config.
+      // In all other cases, we want to load map 0 in the series.
+      g_MapPoolList.GetString(0, mapName, sizeof(mapName));
```
nickdnk, line 111: We might not be able to put 0 here. If someone has `series_score` set in their match config, we shouldn't load map 0, should we? If it starts 1-0, we load map 1, skipping the first one?

nickdnk:
Last commit fixes https://github.com/splewis/get5/issues/728

nickdnk:
Tested everything in here and it seems to work as expected. Tested `clinch_series` and also tested that bo2 still works as expected.

scripting/get5.sp
```diff
@@ -1150,9 +1151,6 @@ public Action Event_MatchOver(Event event, const char[] name, bool dontBroadcast
     // trigger it either.
     Stats_ResetGrenadeContainers();
 
-    // Write backup before series score increments
-    WriteBackup();
```
nickdnk, line 1154: I took this out. Per https://github.com/splewis/get5/issues/822 I don't understand why we do it.

nickdnk:
Now also fixes https://github.com/splewis/get5/issues/213

scripting/get5.sp
```diff
@@ -1650,11 +1679,6 @@ public Action Timer_PostKnife(Handle timer) {
   EnsureIndefiniteWarmup();
 }
 
-public Action StopDemo(Handle timer) {
```
nickdnk, line 1653: This method was not used anywhere.

PhlexPlexico: Couldn't we use this method though as a helper for #381?

nickdnk: We already have a helper. I created a new file called `recording.sp` and put everything GOTV-related into that:

https://github.com/splewis/get5/pull/821/files#diff-bd010f45d65060a909e17e0429128077905e24c818c5d4d9a85cc4cc83702571R1-R24

nickdnk:
Fixes https://github.com/splewis/get5/issues/190 now also.

nickdnk:
I changed the translation logic to always inject the color around variables if applicable. This means we don't have to make sure different languages use the same colors. I also cleaned up some of the translations a bit, where I could.

Player names are now colored based on the side they are on, so for instance when `PlayerName has requested a tactical pause.` is printed, `PlayerName` will now be either yellow or blue, depending on the side the player is on.

Commands in translations (.ie. `type !unpause to continue`) are now always colored green and have been changed to be injected variables as well.

nickdnk:
I changed the translation logic to always inject the color around variables if applicable. This means we don't have to make sure different languages use the same colors. I also cleaned up some of the translations a bit, where I could.

Player names are now colored based on the side they are on, so for instance when `PlayerName has requested a tactical pause.` is printed, `PlayerName` will now be either yellow or blue, depending on the side the player is on.

Commands in translations (.ie. `type !unpause to continue`) are now always colored green and have been changed to be injected variables as well.

scripting/get5.sp
splewis: I don't recall any reason for this; I'm okay with removing it. (Worst case, we discover a bug and can actually document it, but I trust your testing, so I don't expect any issue.)

scripting/get5/recording.sp
```diff
@@ -0,0 +1,124 @@
+stock bool StartRecording() {
```
splewis, line 1: nit: it probably makes more sense for this & StopRecording to be a `public` still (though practically it doesn't really matter). The only value of stock is that the function wouldn't get compiled into the plugin if unused (so it's typically reserved for "library" style files included by many plugins).

scripting/get5/recording.sp
```diff
@@ -0,0 +1,124 @@
...
+  DataPack pack = CreateDataPack();
+  pack.WriteString(matchId);
+  pack.WriteCell(mapNumber);
+  pack.WriteString(demoFileName);
+
+  CreateTimer(3.0, Timer_FireStopRecordingEvent, pack);
+}
+
+public Action Timer_StopGoTVRecording(Handle timer, DataPack pack) {
+
```
splewis, line 69: Superfluous blank line.

scripting/get5/recording.sp
```diff
@@ -0,0 +1,124 @@
...
+  int mapNumber = pack.ReadCell();
+  pack.ReadString(demoFileName, sizeof(demoFileName));
+  delete pack;
+
+  StopRecordingCallback(matchId, mapNumber, demoFileName);
+  return Plugin_Handled;
+}
+
+public Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
+
```
splewis, line 83: Superfluous blank line.

splewis (approved):
Looks good overall, just a few tiny nits that I'd like to see tweaked, feel free to merge it.

PR #826, opened 2022-08-10
Remove CurrentScoreInfoMessage translations as it never varies

nickdnk:
This is kind of self-explanatory.


nickdnk:
Tested, merging.

PR #827, opened 2022-08-12
Clean up Native_Message logic to reduce code repetition

nickdnk:
This takes the three different `Native_MessageToXX` functions and reduces them down to one, also cutting down a lot on calls for `Format`. No changes in functionality; just less repetition and string copy code, which can grow exponentially when the MessageToAll is used in a loop.


nickdnk:
This did not work at all as soon as multiple languages were involved. I will roll this back immediately.

PR #831, opened 2022-08-13
Added surrender system

nickdnk:
This allows you to call `!surrender` or `!gg` when applicable to surrender the map. It uses the built-in logic for actually ending the map.

Planned for 0.11.

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -69,7 +69,6 @@ ConVar g_DamagePrintExcessCvar;
 ConVar g_DamagePrintFormatCvar;
 ConVar g_DemoNameFormatCvar;
 ConVar g_DisplayGotvVetoCvar;
-ConVar g_EndMatchOnEmptyServerCvar;
 ConVar g_EventLogFormatCvar;
 ConVar g_FixedPauseTimeCvar;
 ConVar g_KickClientImmunityCvar;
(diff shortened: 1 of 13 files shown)


scripting/get5.sp
```diff
@@ -757,59 +787,73 @@ public void OnClientSayCommand_Post(int client, const char[] command, const char
...
 public void OnMapStart() {
+  LogDebug("OnMapStart");
+
+  // This is a defensive solution that ensures we don't have lingering surrender-timers. If everyone leaves and a player
+  // then joins the server again, the server may change the map, which triggers this. If this happens, we cannot
+  // recover the game state and must force the series to end.
+  if (g_EndMatchOnEmptyServerTimer != INVALID_HANDLE) {
+    if (g_GameState != Get5State_None) {
+      LogDebug("Triggering surrender timer immediately as map was changed.");
+      TriggerTimer(g_EndMatchOnEmptyServerTimer);
```
nickdnk, line 841: Reminder to myself to test if this is async.

nickdnk: Confirmed to be synchronously calling the timer's callback.

nickdnk:
This is now a breaking change because it removes the `get5_end_match_on_empty_server` cvar, as it makes no sense when this system exists.

PR #841, opened 2022-08-15
Adjust GOTV and cvar restore logic

nickdnk:
So it turns out that the GOTV-related stuff I made in https://github.com/splewis/get5/pull/821 caused some problems, as I had slightly misunderstood how the GOTV delay affected recording of demos. This PR fixes all that, and I've tested it a lot.

I've taken tv_delay parameters out of the sample config files and discouraged people from putting tv_ in their warmup/live, as changing the delay while changing game states messes with the demos. I took out tv_delaymapchange as well, as Get5 automatically makes this adjustment when a map ends. tv_-commands should be added to the server permanently (by some other config) or set in the match config cvar section, so they don't vary between stages.

Added a separate section on GOTV to the docs also.

This also fixes some race-condition related stuff in the Get5_OnDemoFinished event if quickly ending/starting matches, using DataPacks

Accidentally pushed to develop, which was reset; replacement for https://github.com/splewis/get5/pull/839

### DIFF
--- a/cfg/get5/live.cfg
+++ b/cfg/get5/live.cfg
@@ -55,6 +55,4 @@ sv_holiday_mode 0
 sv_talk_enemy_dead 0
 sv_talk_enemy_living 0
 sv_voiceenable 1
-tv_delay 105
-tv_delaymapchange 1
 tv_relayvoice 0
--- a/cfg/get5/warmup.cfg
+++ b/cfg/get5/warmup.cfg
@@ -27,8 +27,5 @@ sv_hibernate_when_empty 0
 sv_infinite_ammo 0
 sv_showimpacts 0
 sv_voiceenable 1
-tv_delay 105
-tv_delaymapchange 1
 tv_relayvoice 0
-
 sv_cheats 0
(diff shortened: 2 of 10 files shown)


deLuXe83:
Hi,
I am now on Plugin version: 0.10.0-972ce8f. Runs smooth!

What I experience is that between mapchange there is a 2 Minute wait-time. `tv_delay` is set to 15 sec.

nickdnk:
@deLuXe83 Hello

What is your `mp_match_restart_delay`? It will only be increased if not long enough to let GOTV finish; it will not be reduced if larger as of this version.

deLuXe83:
> @deLuXe83 Hello
> 
> What is your `mp_match_restart_delay`? It will only be increased if not long enough to let GOTV finish; it will not be reduced if larger as of this version.

'mp_match_restart_delay' was not set at all. 

I will give it a try next text, because I really like this:
> Edit: The idea behind this is that you could now set a fixed 5 or 10 minute break between maps in a tournament environment.
idea!!

nickdnk:
You *are* setting it somewhere, because the default value is 15 seconds.

deLuXe83:
.... not on porpuse :) 

I added it now to my 'warmup.cfg' and will monitor it next time.

nickdnk:
Alright, but semantically it belongs in the `live.cfg`, as this never applies to the warmup period.

Issue #840, opened 2022-08-14
get5.inc cannot be included as a standalone file


splewis:
A plugin simply including get5.inc ([example](https://github.com/splewis/csgo-practice-mode/blob/e2369902bb686205fa7b34038cedce5ab1fd4505/scripting/practicemode.sp#L15)) cannot compile since get5.inc uses files defined elsewhere (`get5/util.sp`):

```
/var/lib/jenkins/workspace/csgo-practice-mode/tmp/addons/sourcemod/scripting/include/get5.inc(170) : error 017: undefined symbol "ConvertGet5SideToStringInJson"
/var/lib/jenkins/workspace/csgo-practice-mode/tmp/addons/sourcemod/scripting/include/get5.inc(170) : error 092: number of arguments does not match definition
/var/lib/jenkins/workspace/csgo-practice-mode/tmp/addons/sourcemod/scripting/include/get5.inc(195) : error 017: undefined symbol "ConvertGameStateToStringInJson"
/var/lib/jenkins/workspace/csgo-practice-mode/tmp/addons/sourcemod/scripting/include/get5.inc(195) : error 092: number of arguments does not match definition
/var/lib/jenkins/workspace/csgo-practice-mode/tmp/addons/sourcemod/scripting/include/get5.inc(296) : error 017: undefined symbol "ConvertGet5SideToStringInJson"
/var/lib/jenkins/workspace/csgo-practice-mode/tmp/addons/sourcemod/scripting/include/get5.inc(296) : error 092: number of arguments does not match definition
/var/lib/jenkins/workspace/csgo-practice-mode/tmp/addons/sourcemod/scripting/include/get5.inc(307) : error 017: undefined symbol "ConvertGet5TeamToStringInJson"
/var/lib/jenkins/workspace/csgo-practice-mode/tmp/addons/sourcemod/scripting/include/get5.inc(307) : error 092: number of arguments does not match definition
/var/lib/jenkins/workspace/csgo-practice-mode/tmp/addons/sourcemod/scripting/include/get5.inc(328) : error 017: undefined symbol "ConvertGet5SideToStringInJson"
/var/lib/jenkins/workspace/csgo-practice-mode/tmp/addons/sourcemod/scripting/include/get5.inc(328) : error 092: number of arguments does not match definition
/var/lib/jenkins/workspace/csgo-practice-mode/tmp/addons/sourcemod/scripting/include/get5.inc(442) : error 017: undefined symbol "ConvertGet5TeamToStringInJson"
[...]

PR #843, opened 2022-08-15
Functions used in get5.inc should be included in that file

nickdnk:
Fixes https://github.com/splewis/get5/issues/840

@splewis Agree on no visibility modifiers required here?


splewis:
I believe these should be `stock` so that they don't get compiled into other plugins if unused.

nickdnk:
https://github.com/splewis/get5/issues/840#issuecomment-1216548403 left comment in the wrong place.

Issue #485, opened 2020-03-31
[Request] Adding date to demos and the ability to create (sub)folders.


DennisF10:
I wondeirng if you could splitt the {TIME} cvar in two

e.g.

get5_time_format = %H_%M = {TIME} 
get5_date_fortmat = %Y-%m-%d = {DATE} 

I'm aware that using {DATE} alone would cause problems without using other cvars but if you also add the ability to create directories by its own we could store demos like /{DATE} /{TIME}_{MAPNAME} which would be really great.

TandelK:
@DennisF10 you can also use {TEAM1} vs {TEAM2} also  in the Demo Name Structure.

DennisF10:
> @DennisF10 you can also use {TEAM1} vs {TEAM2} also in the Demo Name Structure.

I know but if you don't set a teamname everytime it doesn't help with organizing demos.

PhlexPlexico:
So I'm working on getting the sub folders going here, and they will only support the matchID to be replaced. However, it appears that demos already support an array of substituion, `{TIME}` included. 

https://github.com/splewis/get5/blob/fe9db191d30cae9a6ae36f0b410aaa46ea76e0d4/scripting/get5.sp#L1873-L1883

https://github.com/splewis/get5/blob/fe9db191d30cae9a6ae36f0b410aaa46ea76e0d4/scripting/get5/recording.sp#L17

I feel like if you're recording demoes, you should be keen to use the match ID as well, and then use the `{TIME}` substitution in it as well, that should help with sorting at least.

PR #844, opened 2022-08-16
Add Support For Demo Uploads via HTTP & Demo Folder W/ Date Stamps

### LINKED ISSUE #354: [Feature Request] Demos uploaded to central server and tied to match
Any chance we can get Demos to auto-upload to the central server after a match, and have a link for the web page viewer to download said demos?

This is like the #1 requested thing I hear when we do CS:GO stuff.

Please? :)

PhlexPlexico:
This update allows a user to supply a web address and bearer token to authorize against a web server that accepts files from a POST call. This should allow users to setup a simple web server (e.g. using express) to allow a route for demos to be posted to. Docs have been updated to include the new cvars that are available, but some edge case work still needs to be done (such as formatting the address and checking if it contains `https(s)` in the URL already. If no URL is provided, we leave a debug statement to tell that this feature is currently disabled.

This update also includes a new feature of `get5_demo_path` which will allow users to set specific paths for their demos. If the path doesn't exist, get5 will create these paths so users will be able to store demos in a way that can be better sorted. Some logic has been generalized and moved into the util class, as they're being used in more than one place now.

Closes #354 and #485

### DIFF
--- a/documentation/docs/configuration.md
+++ b/documentation/docs/configuration.md
@@ -250,19 +250,15 @@ the [`{MATCHID}`](#tag-matchid) variable, i.e. `backups/{MATCHID}/`.<br>**`Defau
 ####`get5_time_format`
 :   Date and time format string. This determines the [`{TIME}`](#tag-time) tag.<br>**`Default: "%Y-%m-%d_%H-%M-%S"`**
 
+####`get5_date_format`
+:   Date format string. This determines the [`{DATE}`](#tag-date) tag.<br>**`Default: "%Y-%m-%d"`**
+
 !!! danger "Advanced users only"
 
     Do not change the time format unless you know what you are doing. Please always include a component of hours,
     minutes and seconds so that [demo files](#get5_demo_name_format) will not be overwritten. You can find the reference
     for formatting a time string [here](https://cplusplus.com/reference/ctime/strftime/). The default example above
     prints time in this format: `2022-06-12_13-15-45`.
-
-####`get5_demo_name_format`
-:   Format to use for demo files when [recording matches](gotv.md). Do not include a file extension (`.dem` is added
-automatically). If you do not include the [`{TIME}`](#tag-time) tag, you will have
-problems with duplicate files if restoring a game from a backup. Note that the [`{MAPNUMBER}`](#tag-mapnumber)
-variable is not zero-indexed. Set to empty string to disable recording
-demos.<br>**`Default: "{TIME}_{MATCHID}_map{MAPNUMBER}_{MAPNAME}"`**
 
 ####`get5_event_log_format`
 :   Format to write event logs to. Set to empty string to disable writing event logs.<br>**`Default: ""`**
@@ -315,6 +311,41 @@ if [`get5_print_damage`](#get5_print_damage) is disabled.<br>
 ####`get5_knife_cfg`
 :   Config file executed for the knife round, relative to `csgo/cfg`.<br>**`Default: "get5/knife.cfg"`**
 
+## Demos
+
+####`get5_demo_upload_url`
+:   If defined, it is the URL at which a server resides to accept connections to upload demos. Requires
+[SteamWorks](../installation/#steamworks) extension. If no protocol is provided, the plugin will prepend
+`http://` to this value. **`Default: ""`**
+
+####`get5_demo_upload_header_key`
+:   If defined, it is the authorization key that is appended to the HTTP request. Requires
+[SteamWorks](../installation/#steamworks) extension. **`Default: Authorization`**
+
+####`get5_demo_upload_header_value`
+:   If defined, it is the authorization value that is appended to the HTTP request. Requires
+[SteamWorks](../installation/#steamworks) extension. This is where your API *key* would
+be set. **`Default: ""`**
+
+####`get5_demo_delete_after_upload`
+:   Whether or not to delete the demo from the game server after successfully uploading
+it to a web server. Requires [`get5_demo_upload_url`](./#get5_demo_upload_url) to be set.
+**`Default: 0`**
+
+####`get5_demo_path`
+:   The folder of saved [demo files](../gotv), relative to the `csgo` directory. You **can** use
+the [`{MATCHID}`](#tag-matchid) and [`{DATE}`](#tag-date) variable, i.e. `demos/{DATE}/{MATCHID}/`.
+Much like [`get5_backup_path`](./#get5_backup_path), the path must **not** start with a slash, and
+must **end with a slash**.
+**`Default: ""`**
+
+####`get5_demo_name_format`
+:   Format to use for demo files when [recording matches](gotv.md). Do not include a file extension (`.dem` is added
+automatically). If you do not include the [`{TIME}`](#tag-time) tag, you will have
+problems with duplicate files if restoring a game from a backup. Note that the [`{MAPNUMBER}`](#tag-mapnumber)
+variable is not zero-indexed. Set to empty string to disable recording
+demos.<br>**`Default: "{TIME}_{MATCHID}_map{MAPNUMBER}_{MAPNAME}"`**
+
 ## Substitution Variables
 
 ### Match/State Substitutes {: #state-substitutes }
@@ -325,6 +356,9 @@ placeholder strings that will be replaced by meaningful values when printed. Not
 
 ####`{TIME}` {: #tag-time}
 :   The current time, determined by [`get5_time_format`](#get5_time_format).
+
+####`{DATE}` {: #tag-date}
+:   The current date, determined by [`get5_date_format`](#get5_date_format).
 
 ####`{MAPNAME}` {: #tag-mapname}
 :   The pretty-printed name of the map, i.e. **Dust II** for `de_dust2`.
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -64,6 +64,7 @@ ConVar g_AutoLoadConfigCvar;
 ConVar g_AutoReadyActivePlayersCvar;
 ConVar g_BackupSystemEnabledCvar;
 ConVar g_CheckAuthsCvar;
+ConVar g_DateFormatCvar;
 ConVar g_DamagePrintCvar;
 ConVar g_DamagePrintExcessCvar;
 ConVar g_DamagePrintFormatCvar;
@@ -108,6 +109,11 @@ ConVar g_VotesRequiredForSurrenderCvar;
 ConVar g_SurrenderVoteTimeLimitCvar;
 ConVar g_SurrenderCooldownCvar;
 ConVar g_SurrenderTimeToRejoinCvar;
+ConVar g_DemoUploadUrlCvar;
+ConVar g_DemoUploadAuthKeyCvar;
+ConVar g_DemoUploadAuthValueCvar;
+ConVar g_DemoUploadDeleteAfterCvar;
+ConVar g_DemoPathCvar;
 
 // Autoset convars (not meant for users to set)
 ConVar g_GameStateCvar;
@@ -239,6 +245,7 @@ ArrayList g_ChatAliasesCommands;
 
 /** Map-game state not related to the actual gameplay. **/
 char g_DemoFileName[PLATFORM_MAX_PATH];
+char g_DemoFileNameLastStartedUpload[PLATFORM_MAX_PATH];
 bool g_MapChangePending = false;
 bool g_PendingSideSwap = false;
 Handle g_PendingMapChangeTimer = INVALID_HANDLE;
@@ -364,12 +371,30 @@ public void OnPluginStart() {
   g_DamagePrintExcessCvar = CreateConVar(
       "get5_print_damage_excess", "0",
       "Prints full damage given in the damage report on round end. With this disabled (default), a player cannot take more than 100 damage.");
+  g_DateFormatCvar = CreateConVar(
+      "get5_date_format", "%Y-%m-%d",
+      "Date format to use when creating file names. Don't tweak this unless you know what you're doing! Avoid using spaces or colons.");
   g_CheckAuthsCvar =
       CreateConVar("get5_check_auths", "1",
                    "If set to 0, get5 will not force players to the correct team based on steamid");
   g_DemoNameFormatCvar = CreateConVar(
       "get5_demo_name_format", "{TIME}_{MATCHID}_map{MAPNUMBER}_{MAPNAME}",
       "Format for demo file names, use \"\" to disable. Do not remove the {TIME} placeholder if you use the backup system.");
+  g_DemoPathCvar = CreateConVar(
+      "get5_demo_path", "",
+      "The folder to save demo files in, relative to the csgo directory. If defined, it must not start with a slash and must end with a slash.");
+  g_DemoUploadAuthKeyCvar = CreateConVar(
+      "get5_demo_upload_header_value", "",
+      "If defined, it is the authorization value that is appended to the HTTP request. Requires SteamWorks.");
+  g_DemoUploadAuthValueCvar = CreateConVar(
+      "get5_demo_upload_header_key", "Authorization",
+      "If defined, it is the authorization key that is appended to the HTTP request. Requires SteamWorks.");
+  g_DemoUploadDeleteAfterCvar = CreateConVar(
+      "get5_demo_delete_after_upload", "0",
+      "Whether to delete the demo from the game server after a successful upload.");
+  g_DemoUploadUrlCvar = CreateConVar(
+      "get5_demo_upload_url", "",
+      "If defined, it is the URL at which a server resides to accept connections to upload demos. Requires SteamWorks extension. If no protocol is provided, the plugin will prepend http:// to this value.");
   g_DisplayGotvVetoCvar =
       CreateConVar("get5_display_gotv_veto", "0",
                    "Whether to wait for map vetos to be printed to GOTV before changing map");
@@ -1883,10 +1908,14 @@ bool FormatCvarString(ConVar cvar, char[] buffer, int len) {
 
   // Get the time, this is {TIME} in the format string.
   char timeFormat[64];
+  char dateFormat[64];
   g_TimeFormatCvar.GetString(timeFormat, sizeof(timeFormat));
+  g_DateFormatCvar.GetString(dateFormat, sizeof(dateFormat));
   int timeStamp = GetTime();
   char formattedTime[64];
+  char formattedDate[64];
   FormatTime(formattedTime, sizeof(formattedTime), timeFormat, timeStamp);
+  FormatTime(formattedDate, sizeof(formattedDate), dateFormat, timeStamp);
 
   // Get team names with spaces removed.
   char team1Str[MAX_CVAR_LENGTH];
@@ -1899,6 +1928,7 @@ bool FormatCvarString(ConVar cvar, char[] buffer, int len) {
 
   // MATCHTITLE must go first as it can contain other placeholders
   ReplaceString(buffer, len, "{MATCHTITLE}", g_MatchTitle);
+  ReplaceString(buffer, len, "{DATE}", formattedDate);
   ReplaceStringWithInt(buffer, len, "{MAPNUMBER}", Get5_GetMapNumber() + 1);
   ReplaceStringWithInt(buffer, len, "{MAXMAPS}", g_NumberOfMapsInSeries);
   ReplaceString(buffer, len, "{MATCHID}", g_MatchID);
--- a/scripting/get5/backups.sp
+++ b/scripting/get5/backups.sp
@@ -168,21 +168,9 @@ void WriteBackup() {
   }
 
   char folder[PLATFORM_MAX_PATH];
-  g_RoundBackupPathCvar.GetString(folder, sizeof(folder));
-  ReplaceString(folder, sizeof(folder), "{MATCHID}", g_MatchID);
-
-  int backupFolderLength = strlen(folder);
-  if (backupFolderLength > 0 &&
-      (folder[0] == '/' || folder[0] == '.' || folder[backupFolderLength - 1] != '/' ||
-       StrContains(folder, "//") != -1)) {
-    LogError(
-        "get5_backup_path must end with a slash and must not start with a slash or dot. It will be reset to an empty string! Current value: %s",
-        folder);
-    folder = "";
-    g_RoundBackupPathCvar.SetString(folder, false, false);
-  } else {
-    CreateBackupFolderStructure(folder);
-  }
+  char variableSubstitutes[][] = {"{MATCHID}"};
+  CheckAndCreateFolderPath(g_RoundBackupPathCvar, variableSubstitutes, 1, folder,
+                           sizeof(folder));
 
   char path[PLATFORM_MAX_PATH];
   if (g_GameState == Get5State_Live) {
@@ -195,42 +183,6 @@ void WriteBackup() {
   LogDebug("Writing backup to %s", path);
   WriteBackupStructure(path);
   g_LastGet5BackupCvar.SetString(path);
-}
-
-static bool CreateDirectoryWithPermissions(const char[] directory) {
-  LogDebug("Creating directory: %s", directory);
-  return CreateDirectory(directory,  // sets 777 permissions.
-                         FPERM_U_READ | FPERM_U_WRITE | FPERM_U_EXEC | FPERM_G_READ |
-                             FPERM_G_WRITE | FPERM_G_EXEC | FPERM_O_READ | FPERM_O_WRITE |
-                             FPERM_O_EXEC);
-}
-
-static bool CreateBackupFolderStructure(const char[] path) {
-  if (strlen(path) == 0 || DirExists(path)) {
-    return true;
-  }
-
-  LogDebug("Creating backup directory %s because it does not exist.", path);
-  char folders[16][PLATFORM_MAX_PATH];  // {folder1, folder2, etc}
-  char fullFolderPath[PLATFORM_MAX_PATH] =
-      "";  // initially empty, but we append every time a folder is created/verified
-  char currentFolder[PLATFORM_MAX_PATH];  // shorthand for folders[i]
-
-  ExplodeString(path, "/", folders, sizeof(folders), PLATFORM_MAX_PATH, true);
-  for (int i = 0; i < sizeof(folders); i++) {
-    currentFolder = folders[i];
-    if (strlen(currentFolder) ==
-        0) {  // as the loop is a fixed size, we stop when there are no more pieces.
-      break;
-    }
-    // Append the current folder to the full path
-    Format(fullFolderPath, sizeof(fullFolderPath), "%s%s/", fullFolderPath, currentFolder);
-    if (!DirExists(fullFolderPath) && !CreateDirectoryWithPermissions(fullFolderPath)) {
-      LogError("Failed to create or verify existence of directory: %s", fullFolderPath);
-      return false;
-    }
-  }
-  return true;
 }
 
 static void WriteBackupStructure(const char[] path) {
--- a/scripting/get5/debug.sp
+++ b/scripting/get5/debug.sp
@@ -128,6 +128,7 @@ static void AddGlobalStateInfo(File f) {
   f.WriteLine("g_PauseType = %d", g_PauseType);
   f.WriteLine("g_LatestPauseDuration = %d", g_LatestPauseDuration);
   f.WriteLine("g_PendingSurrenderTeam = %d", g_PendingSurrenderTeam);
+  f.WriteLine("g_DemoFileNameLastStartedUpload = %s", g_DemoFileNameLastStartedUpload);
 
   LOOP_TEAMS(team) {
     GetTeamString(team, buffer, sizeof(buffer));
--- a/scripting/get5/recording.sp
+++ b/scripting/get5/recording.sp
@@ -14,20 +14,26 @@ bool StartRecording() {
   }
 
   char demoName[PLATFORM_MAX_PATH + 1];
+
   if (!FormatCvarString(g_DemoNameFormatCvar, demoName, sizeof(demoName))) {
     LogError("Failed to format demo filename. Please check your demo file format convar.");
     g_DemoFileName = "";
     return false;
   }
 
-  FormatEx(g_DemoFileName, sizeof(g_DemoFileName), "%s.dem", demoName);
+  char demoFolder[PLATFORM_MAX_PATH];
+  char variableSubstitutes[][] = {"{MATCHID}", "{DATE}"};
+  CheckAndCreateFolderPath(g_DemoPathCvar, variableSubstitutes, 2, demoFolder,
+                           sizeof(demoFolder));
+
+  char demoPath[PLATFORM_MAX_PATH];
+  FormatEx(demoPath, sizeof(demoPath), "%s%s", demoFolder, demoName);
+  FormatEx(g_DemoFileName, sizeof(g_DemoFileName), "%s%s.dem", demoFolder, demoName);
   LogMessage("Recording to %s", g_DemoFileName);
 
   // Escape unsafe characters and start recording. .dem is appended to the filename automatically.
-  char szDemoName[PLATFORM_MAX_PATH + 1];
-  strcopy(szDemoName, sizeof(szDemoName), demoName);
-  ReplaceString(szDemoName, sizeof(szDemoName), "\"", "\\\"");
-  ServerCommand("tv_record \"%s\"", szDemoName);
+  ReplaceString(demoPath, sizeof(demoPath), "\"", "\\\"");
+  ServerCommand("tv_record \"%s\"", demoPath);
   Stats_SetDemoName(g_DemoFileName);
   return true;
 }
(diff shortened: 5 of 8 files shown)


documentation/docs/configuration.md
```diff
@@ -242,6 +242,14 @@ to disable. **`Default: ""`**
 ####`get5_warmup_cfg`
 :   Config file executed in warmup periods. **`Default: get5/warmup.cfg`**
 
+####`get5_demo_upload_url`
+:   If defined, it is the URL at which a server resides to accept connections to upload demos. Requires SteamWorks
+extension.
+
+####`get5_demo_upload_token`
+:   If defined, it is the authorization token used to upload to an accepting web server. Requires SteamWorks
```
nickdnk, line 250: Explain which kind of authorization token, as you've currently fixed it to `Authorization: Bearer %s`

nickdnk (changes requested):

scripting/get5/recording.sp
```diff
@@ -91,6 +92,51 @@ public Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+  Handle demoRequest = SteamWorks_CreateHTTPRequest(k_EHTTPMethodPOST, formattedUrl);
+
+  if (demoRequest == INVALID_HANDLE) {
+    LogDebug("Failed to create request %s", formattedUrl);
+    return Plugin_Stop;
+  }
+
+  if (!SteamWorks_SetHTTPRequestHeaderValue(demoRequest, "Authorization", demoBearer)) {
+    LogError("Failed to add Bearer token to call. Will end request here.");
+    return Plugin_Stop;
```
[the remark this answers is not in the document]
nickdnk: Edit: Per my last comment, you don't need a timer callback at all.

documentation/docs/configuration.md
nickdnk: It might be better to make the variable "Bearer <token>" so people can use any type of `Authorization` they want.

PhlexPlexico: I think the issue with having different authorization is that it would change the type of header that is sent in. For example, a Bearer token has the http key as "Authorization" where an API key has the http key as "Basic". So we could either do a split on the string that's provided by the cvar and attempt to swap the header key with a case/if statement, or have one set authorization type.

nickdnk: If you want that, you will have to provide 2 cvars. One can just default to `Authorization` then and the token to blank.

nickdnk:
Also, please add that SteamWorks is also required for uploading of demos under that section, so they link to each other:

```
SteamWorks is not required for Get5 to work on your game server, however it is required if you wish to [load match
configs remotely](../commands#get5_loadmatch_url), [automatically
check for updates](../configuration#get5_print_update_notice) or [upload demos](gotv.md).
```

And then add something to the `gotv.md` file about uploading demos and link to the convar again using [].

PhlexPlexico:
Per comments, I've adjusted the code to check and ensure everything is in line with what was suggested.

Also, the only reason this is in draft right now is I have yet to test. I've been meaning to create a very simple web server to ensure this is working. As soon as I can test this, I'll mark it as ready for merge.

Thanks!

scripting/get5/recording.sp
```diff
@@ -94,9 +94,63 @@ public Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+  // Store the filename of the demo for the callback.
+  Format(g_DemoFileNameLastUploaded, sizeof(g_DemoFileNameLastUploaded), demoFileName);
+
+  if (demoRequest == INVALID_HANDLE) {
+    LogDebug("Failed to create request %s", formattedUrl);
+    return;
+  }
+
+  if (!SteamWorks_SetHTTPRequestHeaderValue(demoRequest, demoAuthKey, demoAuthValue)) {
+    LogError("Failed to add auth key and token to call. Will end request here.");
```
nickdnk, line 143: `delete` demoRequest. You should also check that **both** `demoAuthKey` and `demoAuthValue` are defined before you proceed here.

scripting/get5/recording.sp
```diff
@@ -94,9 +94,63 @@ public Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+    return;
+  }
+
+  SteamWorks_SetHTTPRequestRawPostBodyFromFile(demoRequest, "application/octet-stream",
+                                               g_DemoFileNameLastUploaded);
+  SteamWorks_SetHTTPCallbacks(demoRequest, DemoRequestCallback);
+  SteamWorks_SendHTTPRequest(demoRequest);
+  return;
```
nickdnk, line 151: Redundant `return`

scripting/get5/recording.sp
```diff
@@ -136,3 +190,30 @@ void SetCurrentMatchRestartDelay(float delay) {
...
+      if (FileExists(g_DemoFileNameLastUploaded)) {
+        DeleteFile(g_DemoFileNameLastUploaded);
+      }
+
+      // Reset/remove last demo uploaded as we are finished with it.
+      g_DemoFileNameLastUploaded = "";
+    }
+  }
+  delete request;
+  return;
```
nickdnk, line 218: Redundant return

scripting/get5/recording.sp
```diff
@@ -136,3 +190,30 @@ void SetCurrentMatchRestartDelay(float delay) {
...
+  } else {
+    LogDebug("Demo request succeeded, HTTP status code = %d", statusCode);
+    char response[1024];
+    SteamWorks_GetHTTPResponseBodyData(request, response, sizeof(response));
+    LogDebug(response);
+    if (g_DemoUploadDeleteAfterCvar.BoolValue && !StrEqual(g_DemoFileNameLastUploaded, "")) {
+      LogDebug("g_DemoUploadDeleteAfterCvar set to true, deleteing the file from the game server.");
+
+      if (FileExists(g_DemoFileNameLastUploaded)) {
+        DeleteFile(g_DemoFileNameLastUploaded);
```
nickdnk, line 210: Maybe do a `LogError` here if the file deletion fails for whatever reason. `DeleteFile` returns false then.

scripting/get5/recording.sp
```diff
@@ -136,3 +190,30 @@ void SetCurrentMatchRestartDelay(float delay) {
...
+    char response[1024];
+    SteamWorks_GetHTTPResponseBodyData(request, response, sizeof(response));
+    LogError("Response body: %s", response);
+  } else {
+    LogDebug("Demo request succeeded, HTTP status code = %d", statusCode);
+    char response[1024];
+    SteamWorks_GetHTTPResponseBodyData(request, response, sizeof(response));
+    LogDebug(response);
+    if (g_DemoUploadDeleteAfterCvar.BoolValue && !StrEqual(g_DemoFileNameLastUploaded, "")) {
+      LogDebug("g_DemoUploadDeleteAfterCvar set to true, deleteing the file from the game server.");
```
nickdnk, line 207: Typo in `deleting`

scripting/get5/recording.sp
```diff
@@ -136,3 +190,30 @@ void SetCurrentMatchRestartDelay(float delay) {
     mp_match_restart_delay.FloatValue = delay;
   }
 }
+
+public int DemoRequestCallback(Handle request, bool failure, bool requestSuccessful,
+                        EHTTPStatusCode statusCode) {
+  if (failure || !requestSuccessful) {
+    LogError("Demo request failed, HTTP status code = %d", statusCode);
+    char response[1024];
```
nickdnk, line 198: You cannot do this. You don't know the length of the body. Do this:

```c
int responseSize;
SteamWorks_GetHTTPResponseBodySize(request, responseSize);
char[] response = new char[responseSize];
SteamWorks_GetHTTPResponseBodyData(request, response, responseSize);
```

You also have this response code twice, so maybe you can if/else it a bit better.

scripting/get5.sp
```diff
@@ -446,6 +451,22 @@ public void OnPluginStart() {
...
+  g_DemoUploadAuthKeyCvar = CreateConVar(
+      "get5_demo_upload_header_key", "Authorization",
+      "If defined, it is the authorization key that is appended to the HTTP request. Requires SteamWorks extension.");
+
+  g_DemoUploadAuthValueCvar = CreateConVar(
+      "get5_demo_upload_header_value", "",
+      "If defined, it is the authorization value that is appended to the HTTP request. Requires SteamWorks extension.");
+
+  g_DemoUploadDeleteAfterCvar =
+      CreateConVar("get5_demo_upload_delete_after_upload", "0",
```
nickdnk, line 467: Maybe something like `get5_demo_upload_autodelete` is bit better so we don't have `upload` twice

scripting/get5/recording.sp
```diff
@@ -94,9 +94,63 @@ public Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+
+  // Add the protocol strings if missing (only http).
+  if (StrContains(formattedUrl, "http://") == -1 && StrContains(formattedUrl, "https://") == -1) {
+    Format(formattedUrl, sizeof(formattedUrl), "http://%s", formattedUrl);
+  }
+
+  LogDebug("UploadDemoToServer: formattedUrl (SteamWorks) = %s", formattedUrl);
+  Handle demoRequest = SteamWorks_CreateHTTPRequest(k_EHTTPMethodPOST, formattedUrl);
+  // Store the filename of the demo for the callback.
+  Format(g_DemoFileNameLastUploaded, sizeof(g_DemoFileNameLastUploaded), demoFileName);
```
nickdnk, line 135: Move this down to `SteamWorks_SendHTTPRequest` inside an if, so it doesn't get set if the request is not sent - you could also add a LogError there if you want.

```c
if (SteamWorks_SendHTTPRequest(demoRequest)) {
  Format(g_DemoFileNameLastUploaded, sizeof(g_DemoFileNameLastUploaded), demoFileName);
}
```

nickdnk (changes requested):

scripting/get5/recording.sp
```diff
@@ -94,9 +94,63 @@ public Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+
+  UploadDemoToServer(demoFileName);
   return Plugin_Handled;
 }
 
+void UploadDemoToServer(const char[] demoFileName) {
+  bool steamWorksAvaliable = LibraryExists("SteamWorks");
+
+  if (!steamWorksAvaliable) {
+    LogDebug("Cannot upload demoes to a web server without the SteamWorks extension running.");
```
nickdnk, line 106: Also this needs to go **after** the "will not upload because URL is not set, otherwise everyone gets this error all the time if they don't have SteamWorks.

documentation/docs/configuration.md
```diff
@@ -241,6 +241,24 @@ if no match is loaded. Set to empty string to disable. **`Default: ""`**
...
+[SteamWorks](../installation/#steamworks) extension. **`Default: Authorization`**
+
+####`get5_demo_upload_header_value`
+:   If defined, it is the authorization value that is appended to the HTTP request. Requires
+[SteamWorks](../installation/#steamworks) extension. This is where your API *key* would
+be set.
+
+####`get5_demo_upload_autodelete`
+:   Whether or not to delete the demo from the game server after successfully uploading
+it to a web server. Requires [`get5_demo_upload_url`](./#get5_demo_upload_url) to be set.
```
nickdnk, line 259: Must be `../configuration/#get5_demo_upload_url`

scripting/get5/recording.sp
```diff
@@ -94,9 +94,68 @@ public Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+  if (!StrEqual(demoAuthKey, "") && !StrEqual(demoAuthValue, "")) {
+    if (!SteamWorks_SetHTTPRequestHeaderValue(demoRequest, demoAuthKey, demoAuthValue)) {
+      LogError("Failed to add auth key and token to call. Will end request here.");
+      delete demoRequest;
+      return;
+    }
+  }
+
+  SteamWorks_SetHTTPRequestRawPostBodyFromFile(demoRequest, "application/octet-stream",
+                                               g_DemoFileNameLastUploaded);
```
nickdnk, line 149: This should be `demoFileName`, not `g_DemoFileNameLastUploaded`.

scripting/get5/recording.sp
```diff
@@ -94,9 +94,71 @@ public Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+                                                    demoFileName)) {
+    delete demoRequest;
+    return;
+  }
+  SteamWorks_SetHTTPCallbacks(demoRequest, DemoRequestCallback);
+  if (SteamWorks_SendHTTPRequest(demoRequest)) {
+    // Store the filename of the demo for the callback.
+    Format(g_DemoFileNameLastUploaded, sizeof(g_DemoFileNameLastUploaded), demoFileName);
+  } else {
+    LogError("UploadDemoToServer: An error occured when trying to send the HTTP request.");
```
PhlexPlexico, line 158: Do we need to delete `demoRequest` here too, or does the Callback take care of the deletion if the `SteamWorks_SendHTTPRequest` fails?

nickdnk: You must delete it.

nickdnk: I'm not sure that this **can** fail though, to be fair. I think it might only fail if there is a problem with the parameters.

nickdnk: The SteamWorks code looks like this for the call, so it would appear to only fail if `demoRequest` is an invalid handle:

```c
ISteamHTTP *pHTTP;
SteamWorksHTTPRequest *pRequest = GetRequestPointer(pHTTP, pContext, params[1]);
if (pRequest == NULL)
{
	return 0;
}

SteamAPICall_t hCall;
cell_t result = pHTTP->SendHTTPRequest(pRequest->request, &hCall) ? 1 : 0;

SetCallbacks(hCall, pRequest);
return result;
```

 ... which at this point should not be possible.

scripting/get5/recording.sp
```diff
@@ -136,3 +198,32 @@ void SetCurrentMatchRestartDelay(float delay) {
     mp_match_restart_delay.FloatValue = delay;
   }
 }
+
+public int DemoRequestCallback(Handle request, bool failure, bool requestSuccessful,
+                        EHTTPStatusCode statusCode) {
+  int responseSize;
+  SteamWorks_GetHTTPResponseBodySize(request, responseSize);
+  char[] response = new char[responseSize];
+  SteamWorks_GetHTTPResponseBodyData(request, response, responseSize);
```
nickdnk, line 207: When before the `failure` check, you must test if this works for network connection errors, which produce no HTTP response body.

PhlexPlexico: Per SteamWorks docs, it appears we can just check that as well with an `if` - 
> Returns: bool
Returns true upon success indicating that pBodyDataBuffer has been filled with the body data.
Otherwise, returns false under the following conditions:
hRequest was invalid.
The request has not been sent or has not completed.
The request is a streaming request.
pBodyDataBuffer is NULL.

Where pBodyDataBuffer is our `response` variable, if I'm reading this correctly?

nickdnk: Yes, that seems possible.

scripting/get5/recording.sp
```diff
@@ -94,9 +94,84 @@ public Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+  // to an end point that has no authentication.
+  if (!StrEqual(demoAuthKey, "") && !StrEqual(demoAuthValue, "")) {
+    if (!SteamWorks_SetHTTPRequestHeaderValue(demoRequest, demoAuthKey, demoAuthValue)) {
+      LogError("Failed to add auth key and token to call. Will end request here.");
+      delete demoRequest;
+      return;
+    }
+  }
+
+  if (!SteamWorks_SetHTTPRequestHeaderValue(demoRequest, "Content-Disposition", contentFilename)) {
```
nickdnk, line 152: This, unfortunately won't work if we are to keep within the HTTP spec. It only applies when using `multipart/form-data`.

We should go back to the `Get5-Demoname` header solution.

scripting/get5/recording.sp
```diff
@@ -94,9 +94,84 @@ public Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+    LogError("Failed to add filename content header to call. Will end request here.");
+    delete demoRequest;
+    return;
+  }
+
+  if (!SteamWorks_SetHTTPRequestHeaderValue(demoRequest, "User-Agent", userAgent)) {
```
nickdnk, line 158: This does not work. We must use `SteamWorks_SetHTTPRequestUserAgentInfo` and just append our `userAgent` string to the built-in SteamWorks header. That works fine.

nickdnk (changes requested):
Everything else seems to be good. File gets uploaded, auth headers work etc. I have not tested auto-deletion.

documentation/docs/configuration.md
```diff
@@ -241,6 +241,25 @@ if no match is loaded. Set to empty string to disable. **`Default: ""`**
...
+####`get5_demo_upload_header_key`
+:   If defined, it is the authorization key that is appended to the HTTP request. Requires
+[SteamWorks](../installation/#steamworks) extension. **`Default: Authorization`**
+
+####`get5_demo_upload_header_value`
+:   If defined, it is the authorization value that is appended to the HTTP request. Requires
+[SteamWorks](../installation/#steamworks) extension. This is where your API *key* would
+be set.
+
+####`get5_demo_upload_delete_after_upload`
```
nickdnk, line 258: Missing **`Default: ""`**

documentation/docs/configuration.md
```diff
@@ -241,6 +241,25 @@ if no match is loaded. Set to empty string to disable. **`Default: ""`**
...
+####`get5_demo_upload_url`
+:   If defined, it is the URL at which a server resides to accept connections to upload demos. Requires
+[SteamWorks](../installation/#steamworks) extension. If no protocol is provided, the plugin will prepend
+`http://` to this value.
+
+####`get5_demo_upload_header_key`
+:   If defined, it is the authorization key that is appended to the HTTP request. Requires
+[SteamWorks](../installation/#steamworks) extension. **`Default: Authorization`**
+
+####`get5_demo_upload_header_value`
```
nickdnk, line 253: Missing **`Default: ""`**

documentation/docs/configuration.md
```diff
@@ -241,6 +241,25 @@ if no match is loaded. Set to empty string to disable. **`Default: ""`**
 ####`get5_warmup_cfg`
 :   Config file executed in warmup periods. **`Default: get5/warmup.cfg`**
 
+####`get5_demo_upload_url`
```
nickdnk, line 244: Missing **`Default: ""`**

documentation/docs/configuration.md
PhlexPlexico: 0 means disabled. Much like all the other bool values they are default to 0.

nickdnk: Ah yes of course this was a bool. I’m ok my phone and got them mixed up.

nickdnk:
Should we put https://github.com/splewis/get5/issues/485 in this while we're at it?

PhlexPlexico:
Per Discord, I'm a little bit occupied for the next little bit here, so I think we should implement this for 0.11, and then get the time stamp + demo locations in on a different merge. I'll fix up this PR and once 0.10 is out, I can remove the draft status on it.

nickdnk:
I think maybe `g_DemoFileNameLastUploaded` should be renamed to `g_DemoFileNameLastStartedUpload`, as it's reset on completion, so debugging will output a blank string which might be a little misleading if you just uploaded a demo? Just a thought. And also, the variable should be added to the debug output in `debug.sp`.

scripting/get5/debug.sp
```diff
@@ -121,6 +121,7 @@ static void AddGlobalStateInfo(File f) {
   f.WriteLine("g_PausingTeam = %d", g_PausingTeam);
   f.WriteLine("g_PauseType = %d", g_PauseType);
   f.WriteLine("g_LatestPauseDuration = %d", g_LatestPauseDuration);
+  f.WriteLine("g_DemoFileNameLastStartedUpload = %d", g_DemoFileNameLastStartedUpload);
```
nickdnk, line 124: It's a string, so it must be `%s` :)

PhlexPlexico: Hah, my bad 🤐, should be good now.

nickdnk:
> I think resetting it may still have its uses, as we wouldn't want state from a previous game on match load, correct? I'm just following the same process that some of the other globals that are reset on match completion (like matchID as an example).

We could reset it where we reset all the other state; when ending or loading a match. It was more of a way to actually use the debug output, as you'll have to run `get5_debuginfo` between start and end of the request for it to ever have anything in it, currently. It's only relevant for debugging, so it might be overengineering it.

PhlexPlexico:
True, I can see this coming to bite us in the ass when someone reports their demos not uploading as well. 

I'll move it into `LoadMatchConfig` to reset to a blank value. I think that would make the most sense, and it will just reformat itself on every demo completion (which should be every map finish at least), that way we keep track of the last known string until a new series is created.

scripting/get5/recording.sp
```diff
@@ -94,9 +94,84 @@ public Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+
+  if (!SteamWorks_SetHTTPRequestUserAgentInfo(demoRequest, userAgent)) {
+    LogError("Failed to add user-agent header to call. Will end request here.");
+    delete demoRequest;
+    return;
+  }
+
+  if (!SteamWorks_SetHTTPRequestRawPostBodyFromFile(demoRequest, "application/octet-stream",
+                                                    demoFileName)) {
+    delete demoRequest;
```
nickdnk, line 166: Maybe a LogError here? Like if the file suddenly does not exist or cannot be read due to file permissions or whatever.

scripting/get5.sp
```diff
@@ -360,8 +362,15 @@ public void OnPluginStart() {
...
                    "If set to 0, get5 will not force players to the correct team based on steamid");
+  g_DateFormatCvar = CreateConVar(
+      "get5_date_format", "%Y-%m-%d",
+      "Date format to use when creating file names. Don't tweak this unless you know what you're doing! Avoid using spaces or colons.");
+  
   g_DemoNameFormatCvar = CreateConVar("get5_demo_name_format", "{MATCHID}_map{MAPNUMBER}_{MAPNAME}",
                                       "Format for demo file names, use \"\" to disable");
+  g_DemoPathCvar =
+      CreateConVar("get5_demo_path", "",
+                   "Path to save demos to, relative to the csgo directory, use \"\" to disable");
```
nickdnk, line 373: Empty string does not disable, it just puts them in root, no?

scripting/get5/recording.sp
```diff
@@ -14,18 +14,45 @@ bool StartRecording() {
...
+  g_DateFormatCvar.GetString(dateFormat, sizeof(dateFormat));
+  int timeStamp = GetTime();
+  FormatTime(formattedDate, sizeof(formattedDate), dateFormat, timeStamp);
+
+  ReplaceString(demoFolder, sizeof(demoFolder), "{MATCHID}", g_MatchID);
+  ReplaceString(demoFolder, sizeof(demoFolder), "{DATE}", formattedDate, false);
+
+  int demoFolderLength = strlen(demoFolder);
+
+  if (demoFolderLength > 0 &&
```
nickdnk, line 37: Turn this into a utility function as it's a duplicate of the logic used for backups.

nickdnk (changes requested):

scripting/get5.sp
```diff
@@ -1504,18 +1482,7 @@ public void WriteBackup() {
...
-       StrContains(folder, "//") != -1)) {
-    LogError(
-        "get5_backup_path must end with a slash and must not start with a slash or dot. It will be reset to an empty string! Current value: %s",
-        folder);
-    folder = "";
-    g_RoundBackupPathCvar.SetString(folder, false, false);
-  } else {
-    CreateBackupFolderStructure(folder);
-  }
+  CheckAndCreateFolderPath(g_RoundBackupPathCvar);
```
nickdnk, line 1485: You're replacing `{MATCHID}` in `folder` but then you pass the Cvar. You won't have replacement then.

scripting/get5/recording.sp
PhlexPlexico: Moved and created new util function.

scripting/get5.sp
PhlexPlexico: Definition updated to be in line with backup_path.

documentation/docs/configuration.md
```diff
@@ -276,6 +285,34 @@ the [`{MATCHID}`](#tag-matchid) variable, i.e. `backups/{MATCHID}/`. **`Default:
...
+be set. **`Default: ""`**
+
+####`get5_demo_upload_delete_after_upload`
+:   Whether or not to delete the demo from the game server after successfully uploading
+it to a web server. Requires [`get5_demo_upload_url`](./#get5_demo_upload_url) to be set.
+**`Default: 0`**
+
+####`get5_demo_path`
+:   The folder of saved [demo files](../gotv), relative to the `csgo` directory. You **can** use
+the [`{MATCHID}`](#tag-matchid) and [`{DATE}`](#tag-date) variable, i.e. `demoes/{DATE}/{MATCHID}`.
```
nickdnk, line 311: Example should have the required trailing slash.

documentation/docs/configuration.md
```diff
@@ -216,10 +216,19 @@ must confirm. **`Default: 0`**
...
+    In order to adjust these parameters, you will need to know about time formats. SourceMod adheres
+    to `strftime` in C, and the format specifiers can be found [here](https://cplusplus.com/reference/ctime/strftime/). **Please** avoid using spaces or colons in these parameters.
+
 ####`get5_time_format`
 :   Time format string. This determines the [`{TIME}`](#tag-time) tag. **Do not change this unless you know what you are
 doing! Avoid using spaces or colons.** **`Default: %Y-%m-%d_%H`**
 
+####`get5_date_format`
+:   Date format string. This determines the [`{DATE}`](#tag-date) tag. **Do not change this unless you know what you are
+doing! Avoid using spaces or colons.** **`Default: %Y-%m-%d`**
```
nickdnk, line 230: Defaults should be wrapped in quotes.

documentation/docs/configuration.md
```diff
@@ -216,10 +216,19 @@ must confirm. **`Default: 0`**
...
 **Note: for these, setting the cvar to an empty string ("") will disable the file writing entirely.**
 
+!!! info "Information on Time ⌚"
+
+    In order to adjust these parameters, you will need to know about time formats. SourceMod adheres
+    to `strftime` in C, and the format specifiers can be found [here](https://cplusplus.com/reference/ctime/strftime/). **Please** avoid using spaces or colons in these parameters.
+
 ####`get5_time_format`
 :   Time format string. This determines the [`{TIME}`](#tag-time) tag. **Do not change this unless you know what you are
 doing! Avoid using spaces or colons.** **`Default: %Y-%m-%d_%H`**
```
nickdnk, line 226: Also quotes here.

scripting/get5.sp
```diff
@@ -1503,22 +1481,11 @@ public void WriteBackup() {
...
-       StrContains(folder, "//") != -1)) {
-    LogError(
-        "get5_backup_path must end with a slash and must not start with a slash or dot. It will be reset to an empty string! Current value: %s",
-        folder);
-    folder = "";
-    g_RoundBackupPathCvar.SetString(folder, false, false);
-  } else {
-    CreateBackupFolderStructure(folder);
-  }
+  folder = CheckAndCreateFolderPath(g_RoundBackupPathCvar, variableSubstitutes, variableListSize);
```
nickdnk, line 1488: To follow the standard we normally use for strings, you should probably provide a buffer and buffer size argument to the function instead of returning a `char[]`.

scripting/get5/recording.sp
```diff
@@ -14,20 +14,26 @@ bool StartRecording() {
...
     LogError("Failed to format demo filename. Please check your demo file format convar.");
     g_DemoFileName = "";
     return false;
   }
 
-  Format(g_DemoFileName, sizeof(g_DemoFileName), "%s.dem", demoName);
+  char demoFolder[PLATFORM_MAX_PATH + 1];
+  char variableSubstitutes[][] = {"{MATCHID}", "{DATE}"};
+  int variableListSize = sizeof(variableSubstitutes);
+  demoFolder = CheckAndCreateFolderPath(g_DemoPathCvar, variableSubstitutes, variableListSize);
```
nickdnk, line 27: Also here, the return vs buffer pattern.

scripting/get5/util.sp
```diff
@@ -662,6 +662,79 @@ public bool IsJSONPath(const char[] path) {
...
+stock char[] CheckAndCreateFolderPath(const ConVar cvar, const char[][] varsToReplace,
+                                      const int varListSize) {
+  char path[PLATFORM_MAX_PATH + 1];
+  char cvarName[MAX_CVAR_LENGTH];
+
+  cvar.GetName(cvarName, sizeof(cvarName));
+  cvar.GetString(path, sizeof(path));
+
+  for (int i = 0; i < varListSize; i++) {
+    if (StrEqual("{MATCHID}", varsToReplace[i])) {
```
nickdnk, line 710: Missing case sensitivity here. Both compares and replaces should have the false argument.

nickdnk (changes requested):

nickdnk (approved):

scripting/get5.sp
```diff
@@ -1894,6 +1923,7 @@ bool FormatCvarString(ConVar cvar, char[] buffer, int len) {
   ReplaceString(team2Str, sizeof(team2Str), " ", "_");
 
   // MATCHTITLE must go first as it can contain other placeholders
+  ReplaceString(buffer, len, "{DATE}", formattedDate, false);
```
nickdnk, line 1926: Remove all case insensitivity for placeholders. I took it away in a PR on dev, as you can see in the calls below.

scripting/get5/recording.sp
```diff
@@ -95,9 +102,91 @@ static Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+    LogDebug("Cannot upload demoes to a web server without the SteamWorks extension running.");
+    return;
+  }
+
+  char formattedUrl[1024];
+  char contentFilename[1024];
+  char userAgent[1024];
+  strcopy(formattedUrl, sizeof(formattedUrl), demoUrl);
+  ReplaceString(formattedUrl, sizeof(formattedUrl), "\"", "");
+  Format(contentFilename, sizeof(contentFilename), "attachment;filename=%s", demoFileName);
```
nickdnk, line 134: Remove `contentFilename` entirely.

scripting/get5/recording.sp
```diff
@@ -137,3 +226,35 @@ void SetCurrentMatchRestartDelay(float delay) {
     mp_match_restart_delay.FloatValue = delay;
   }
 }
+
+static int DemoRequestCallback(Handle request, bool failure, bool requestSuccessful,
```
nickdnk, line 230: Make the callback function this. Note that if you rename `get5_demo_upload_delete_after_upload`, also replace it here. I tested this code and it works for success, HTTP errors and network errors.

```c
static int DemoRequestCallback(Handle request, bool failure, bool requestSuccessful,
                               EHTTPStatusCode statusCode) {
  if (failure || !requestSuccessful) {
    LogError("Failed to upload demo %s.", g_DemoFileNameLastStartedUpload);
    delete request;
    return;
  }
  int status = view_as<int>(statusCode);
  if (status >= 300 || status < 200) {
    LogError("Demo request failed with HTTP status code: %d.", statusCode);
    int responseSize;
    SteamWorks_GetHTTPResponseBodySize(request, responseSize);
    char[] response = new char[responseSize];
    if (SteamWorks_GetHTTPResponseBodyData(request, response, responseSize)) {
      LogError("Response body: %s", response);
    } else {
      LogError("Failed to read response body.");
    }
    delete request;
    return;
  }
  LogDebug("Demo request succeeded. HTTP status code: %d.", statusCode);
  if (g_DemoUploadDeleteAfterCvar.BoolValue && !StrEqual(g_DemoFileNameLastStartedUpload, "")) {
    LogDebug("get5_demo_upload_delete_after_upload set to true; deleting the file from the game server.");
    if (FileExists(g_DemoFileNameLastStartedUpload)) {
      if (!DeleteFile(g_DemoFileNameLastStartedUpload)) {
        LogError("Unable to delete demo file %s.", g_DemoFileNameLastStartedUpload);
      }
    }
  }
  delete request;
}
```

scripting/get5/recording.sp
```diff
@@ -14,20 +14,27 @@ bool StartRecording() {
...
   if (!FormatCvarString(g_DemoNameFormatCvar, demoName, sizeof(demoName))) {
     LogError("Failed to format demo filename. Please check your demo file format convar.");
     g_DemoFileName = "";
     return false;
   }
 
-  FormatEx(g_DemoFileName, sizeof(g_DemoFileName), "%s.dem", demoName);
+  char demoFolder[PLATFORM_MAX_PATH];
+  char variableSubstitutes[][] = {"{MATCHID}", "{DATE}"};
+  int variableListSize = sizeof(variableSubstitutes);
```
nickdnk, line 26: You can replace this with inline `2`, or just inline the `sizeof()` call at least. The `variableListSize` variable is redundant.

```c
CheckAndCreateFolderPath(g_DemoPathCvar, variableSubstitutes, 2, demoFolder,
                           sizeof(demoFolder));
```

scripting/get5/util.sp
```diff
@@ -670,6 +670,80 @@ stock bool IsJSONPath(const char[] path) {
...
+                                    const int varListSize, char outputFolder[PLATFORM_MAX_PATH],
+                                    const int outputFolderSize) {
+  char path[PLATFORM_MAX_PATH];
+  char cvarName[MAX_CVAR_LENGTH];
+
+  cvar.GetName(cvarName, sizeof(cvarName));
+  cvar.GetString(path, sizeof(path));
+
+  for (int i = 0; i < varListSize; i++) {
+    if (StrEqual("{MATCHID}", varsToReplace[i], false)) {
```
nickdnk, line 719: Remove case insensitivity here also, for `ReplaceString` and `StrEqual` calls.

scripting/get5/recording.sp
```diff
@@ -95,9 +102,91 @@ static Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+  // to an end point that has no authentication.
+  if (!StrEqual(demoAuthKey, "") && !StrEqual(demoAuthValue, "")) {
+    if (!SteamWorks_SetHTTPRequestHeaderValue(demoRequest, demoAuthKey, demoAuthValue)) {
+      LogError("Failed to add auth key and token to call. Will end request here.");
+      delete demoRequest;
+      return;
+    }
+  }
+
+  if (!SteamWorks_SetHTTPRequestHeaderValue(demoRequest, "Get5-Demoname", contentFilename)) {
```
nickdnk, line 160: Should be `demoFileName` instead of `contentFilename`.

documentation/docs/configuration.md
```diff
@@ -315,6 +318,38 @@ if [`get5_print_damage`](#get5_print_damage) is disabled.<br>
...
+it to a web server. Requires [`get5_demo_upload_url`](./#get5_demo_upload_url) to be set.
+**`Default: 0`**
+
+####`get5_demo_path`
+:   The folder of saved [demo files](../gotv), relative to the `csgo` directory. You **can** use
+the [`{MATCHID}`](#tag-matchid) and [`{DATE}`](#tag-date) variable, i.e. `demoes/{DATE}/{MATCHID}/`.
+Much like [`get5_backup_path`](./#get5_backup_path), the path must **not** start with a slash, and
+must **end with a slash**.
+**`Default: ""`**
+
```
nickdnk, line 352: Move `get5_demo_name_format` docs down here, so we group all demo-related stuff. It's above under `Formats & Paths` currently.

nickdnk (changes requested):

documentation/docs/configuration.md
```diff
@@ -315,6 +318,38 @@ if [`get5_print_damage`](#get5_print_damage) is disabled.<br>
...
+####`get5_demo_upload_header_key`
+:   If defined, it is the authorization key that is appended to the HTTP request. Requires
+[SteamWorks](../installation/#steamworks) extension. **`Default: Authorization`**
+
+####`get5_demo_upload_header_value`
+:   If defined, it is the authorization value that is appended to the HTTP request. Requires
+[SteamWorks](../installation/#steamworks) extension. This is where your API *key* would
+be set. **`Default: ""`**
+
+####`get5_demo_upload_delete_after_upload`
```
nickdnk, line 341: And yeah rename this to `get5_demo_delete_after_upload`

documentation/docs/configuration.md
```diff
@@ -315,6 +318,38 @@ if [`get5_print_damage`](#get5_print_damage) is disabled.<br>
 ####`get5_knife_cfg`
 :   Config file executed for the knife round, relative to `csgo/cfg`.<br>**`Default: "get5/knife.cfg"`**
 
+####`get5_autoload_config`
```
nickdnk, line 321: Remove `get5_autoload_config` - it's up under server config now. You have it in here twice when adding it here.

scripting/get5/backups.sp
```diff
@@ -168,21 +168,10 @@ void WriteBackup() {
...
-    LogError(
-        "get5_backup_path must end with a slash and must not start with a slash or dot. It will be reset to an empty string! Current value: %s",
-        folder);
-    folder = "";
-    g_RoundBackupPathCvar.SetString(folder, false, false);
-  } else {
-    CreateBackupFolderStructure(folder);
-  }
+  char variableSubstitutes[][] = {"{MATCHID}"};
+  int variableListSize = sizeof(variableSubstitutes);
```
nickdnk, line 172: `variableListSize` again here. Just inline the `sizeof`

scripting/get5/recording.sp
```diff
@@ -95,9 +101,101 @@ static Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+  }
+  
+  if (!SteamWorks_SetHTTPRequestHeaderValue(demoRequest, "Get5-MatchId", matchId)) {
+    LogError("Failed to add filename content header to call. Will end request here.");
+    delete demoRequest;
+    return;
+  }
+  
+  if (!SteamWorks_SetHTTPRequestHeaderValue(demoRequest, "Get5-MapNumber", strMapNumber)) {
+    LogError("Failed to add filename content header to call. Will end request here.");
```
nickdnk, line 176: mapnumber, not filename

scripting/get5/recording.sp
```diff
@@ -95,9 +101,101 @@ static Action Timer_FireStopRecordingEvent(Handle timer, DataPack pack) {
...
+  }
+
+  if (!SteamWorks_SetHTTPRequestUserAgentInfo(demoRequest, userAgent)) {
+    LogError("Failed to add user-agent header to call. Will end request here.");
+    delete demoRequest;
+    return;
+  }
+  
+  if (!SteamWorks_SetHTTPRequestHeaderValue(demoRequest, "Get5-MatchId", matchId)) {
+    LogError("Failed to add filename content header to call. Will end request here.");
```
nickdnk, line 170: matchid, not filename

Sh3nl0ng:
Really great work guys.

PR #849, opened 2022-08-19
Ensure that g_MatchID is reset 

nickdnk:
If you end the series or if a match configuration fails to load, you will be stuck with an invalid `g_MatchId` variable. This causes the backup system to incorrectly look for only backups belonging to that match, where it should be listing all backups.


nickdnk:
@PhlexPlexico Fairly sure we can just merge this immediately, I just wanted a second opinion here.

PhlexPlexico (approved):
Good catch! I agree!

PR #851, opened 2022-08-20
Allow for customizing team color names

nickdnk:
Pretty straight-forward this one.

I reverted the change from https://github.com/splewis/get5/pull/835 as I had not realized the names would be set later during ready-phase, when people had joined.


nickdnk:
Tested. Works fine.

PhlexPlexico (approved):

PR #853, opened 2022-08-21
Empty cvar strings not working in scrim template

nickdnk:
So this one took a while to figure out. Apparently, `KeyValue` does not allow you to write a key with an emptry string as a value, so when we were reading the scrim template, we would potentially write empty strings in there back to the match config file for loading, which simply does nothing, so stuff like `"get5_demo_format" ""` would just silently fail and recording would then not be disabled.

I refactored a bit of code while I was at it, but this fixes it.


scripting/get5/util.sp
```diff
@@ -422,6 +421,26 @@ stock bool RemoveStringFromArray(ArrayList list, const char[] str) {
...
+  kv.GetString(NULL_STRING, buffer, bufferSize);
+  if (StrEqual(KEYVALUE_STRING_PLACEHOLDER, buffer)) {
+    Format(buffer, bufferSize, "");
+    return true;
+  }
+  return false;
+}
+
+stock bool WritePlaceholderInsteadOfEmptyString(const KeyValues kv, char[] buffer, const int bufferSize) {
+  kv.GetString(NULL_STRING, buffer, bufferSize);
```
nickdnk, line 436: Technically this buffer could be inside the function and the arguments could be removed, but since it's always being used in a loop, I let it be a parameter so we don't redefine a `char[]` array each time for no reason.

PR #854, opened 2022-08-22
Reset votes to restore to start of round when a new backup is written

nickdnk:
Thoughts? Ideally we would store the round number the initial command was given during and reset it on a timer instead.


PhlexPlexico:
I think the usage of timers may be over-engineering the practice of the .stop command. Most TOs also have a rule that states restores are not allowed if damage has been taken on either side, so restoring to a point in time when the initial command was given wouldn't make much sense. Worst case scenario, users can list the backups and get an admin to restore at that point in time if needed.

Otherwise, resetting the vote at each round does make sense, but maybe we should have a message output to the team that their vote was reset at the round start, if they had called stop prior to?

nickdnk:
I agree with both. I will add a “stop command vote reset” message if any team has a cast vote when it is reset.

PR #858, opened 2022-08-26
Release 0.10

nickdnk:
These are all changes between 0.9.0 and 0.10.0. I'll update this comment to go over it all.

### I know this is almost impossible to review from start to end, as there is simply so much in it, but I've tested this **a lot** and fixed everything I wanted to fix.

----
Some key points:

1. The `coach` command in console can no longer be used. We simply block it and tell the client to use the `!coach` chat command or `sm_coach` in console. We do this because the built-in coaching mechanisms are hard to work around and cause weird bugs when used in backups. You can only use this command in freezetime and during warmup.
2. You should now **never** set `players_per_team` to more than the number of players you want to actually have playing. In "scrim mode", anyone joining when the team is full will automatically be placed on coach or kicked if there are no more coach slots. If you are coach and not the last to join, simply call `!coach` in chat and you will be moved there. In non-scrim mode - that is, when both teams are provided in a regular match config - anyone in the `coaches` array will *never* be allowed to join the game as anything but a coach. If you are in the regular `players` array and call `!coach`. you will be moved to the coach array and locked there unless removed using `get5_removeplayer`.
3. You can rejoin the game during warmup if you were not supposed to be coaching. Just call `.coach` again. If the team is full, you cannot rejoin the game.
4. If you join the game during halftime, you are now automatically placed on the right team when the following round starts. This applies to both players and coaches and prevents coaches from simply not selecting a team and ghosting.
5. If your `players_per_team` is set to less than the number of players already on a side, and you load the match configuration, the entire team that exceeded the `players_per_team` will be kicked and must rejoin, but this is a caveat I think we can live with.
6. Dedicated documentation added for coaching to explain all this.
7. Added some defensive if-statements when using `get5_addplayer`,  `get5_addkickedplayer` and `get5_addcoach` to prevent these from being executed during backups or halftime.
8. We **only** update the backup on round start, not when adding or removing players or coaches. A backup should be point-in-time and should not be partially modified at a later stage. If you add a coach or player and restore to a game state where this was not done, they should be kicked from the game and the command should be run again. An exception is the native `SetMatchId` which **has** to update the backup as it changes the match ID.

Part 2:
1. Refactored the entire team placement logic. It was a bit of a mess, and this was only because coaching was also a mess.
2. Added a "discouraged parameters" section to the configuration file documentation, as we want to prevent users from putting warmup and restarts into their config files.
3. Got rid of `g_BO2Match` and replace its logic with the added `g_NumberOfMapsInSeries` from https://github.com/splewis/get5/pull/856
4. Fixed a **bunch** of problems with backups, such as being on the wrong side if restoring to a different map and a round number after halftime.
5. Merged `OnConfigsExecuted` and `OnMapStart` since they are always called in succession and SourceMod recommends only using `OnConfigsExecuted` if you set/read cvars: https://sm.alliedmods.net/new-api/sourcemod/OnConfigsExecuted
6. Removed a bunch of `stock` and `public` modifiers.
7. Moved `WriteBackup()` call to `RoundStart` instead of `RoundPreStart`, as it allows for the Valve backup to be written for the first round. For some odd reason, the first round Valve backup has not been written to disk when this call is in `RoundPreStart`, leading to weird handling of valve backups if you wanted to restore to round 0.
8. Added a game restart post-knife, as this is necessary to clear the scoreboard UI. Otherwise the team winning the knife-round will have their win lingering in the middle of the round-win-indicators until the first round is played out. This also helps cleaning up the "going live" logic and tightly couples the announcement to the round start instead of using a timer.
9. Removed a lot of redundant calls to various functions that were stacked on top of eachother.
10. Prevent loading backups during halftime.
11. Added a `RestartGame()` call before loading a Valve backup, as it ensures the game is not in a state that's not prepared for backup (such as bots only alive or after-round).
12. Added pauses used to the backups. They only reset to the backup state if the game is *not* being restored from a live state, meaning they can now be carried over to an entirely new server and restored in a "best effort" manner.
13. We now delete the temporary valve backup file created when exporting the valve backup from a get5 backup.
14. Added `g_MapSides` content to debug output.
15. Fixed missing deletion of "knife decision" timer when a team makes a decision. Previously it would just run its trigger and do nothing because the game was not in a state that expected a knife decision.
16. Fixed some problems with `{MAXMAPS}` and `{MAPNUMBER}` being wrong when formatting them, such as not taking ties or `g_SeriesCanClinch` into consideration.
17. Tightly couple `ExecCfg()` to a call to `SetMatchTeamCvars()` and `ExecuteMatchConfigCvars`, as `ExecCfg` makes a call to `exec xx.cfg` which is async: the two latter must always be set **after** a config file load.
18. Converted `MaxMapsToPlay()` to a more sensible `MapsToWin()` function. The logic was kind of inverted and unnecessary.
19. Fixed a bunch of redundant calls and weird logic in `EnsureIndefiniteWarmup` and merged it into `StartWarmup`.
20. Team-join menu is now never used for any reason.
21. If you disable coaching with `sv_coaching_enabled 0`, existing coaches will be kicked, as they would otherwise become regular spectators. 
22. Added collision protection to the `get5_names.txt` file using the already established logic for this using `{SERVERID}`.


nickdnk:
A few things I've found and corrected while testing this:

1. During backups, the `Get5_OnRoundStart` forward was called twice. This has been corrected so it's suppressed when doing restores and fired explicitly by the backup process and only once.

2. Coaches were being added to the KeyValue stats and had their `STAT_ROUNDSPLAYED` property incremented. This has been corrected, and a `STAT_COACHING` number/bool has been added to the KeyValue stats so they can be filtered out but are still included: https://github.com/splewis/get5/pull/858/commits/84e5241a5d97a475ace3ee66f7627ed5dcd21861#diff-c4154d12ae064f557e3b359f5861f53f7f998ea3945efa06c75eda81429af374R238-R240

nickdnk:
Some more adjustments:

Fixed a problem with lingering global references to map-veto menu.
No longer taking backups during veto/preveto, as it makes no sense and complicates the backup flow a lot.

nickdnk:
Some more findings and fixes:

Cancel and restart the countdown if the match is canceled and make sure no menu events are applied if the state is not veto: https://github.com/splewis/get5/pull/858/commits/7980c87e283eec306f14ef3bc676ac8b41bf542a

Unpause the game if veto is aborted: https://github.com/splewis/get5/pull/858/commits/749febe6eaf3721ef529765a668df5f9cd1fd86e

Put `OnConfigsExecuted`'s trigger of `CheckAutoLoadConfig` on a callback, as this was not working if a server was hibernating: https://github.com/splewis/get5/pull/858/commits/49cee157804fbd7ad8da01c145de68bfdd59faca

Removed a bunch of redundant calls to `IsClientConnected` as this is already established when using `IsClientInGame`: https://github.com/splewis/get5/pull/858/commits/3fc38ed271e4becdd5364dac2d67fa3d512ba277 (see #9 here: https://forums.alliedmods.net/showthread.php?t=321089#9)

Move the `[READY]` / `[NOT READY]` tag **behind** the team name to avoid the UI spectator "bug" where team names are both `[` instead of their starting letter: https://github.com/splewis/get5/pull/858/commits/cb98d19f783d920d4d077a3c47099b57547ee1c5

Inject team into `FormatPlayerName` to always have spectators' and coaches'  names correctly colored: https://github.com/splewis/get5/pull/858/commits/750a81eed5a1d6ae144b30888e70c7fe0e122ec2

Don't replace color tags with the `{WHITE}` color tags when stripping colors for console, just remove the color modifier. This prevents garbled icons in console log output, i.e. `[Get5] Astralis is winning the series 1-0.`: https://github.com/splewis/get5/pull/858/commits/e1ce0c96fc4d49db6a6afe6041b94c55b8b25421

splewis:
I wasn't able to a as-thorough-as-I'd-like review, but LGTM. I do particularly appreciate the documentation about cvars you *really shouldn't set*, as the past was really just me hoping nobody would try to set them :)

splewis (approved):

PR #860, opened 2022-09-02
Format and commit code on push to master branch

PhlexPlexico:
This PR is to just ensure we have correct formatting whenever we do a full release. This little bit of automation here is done so that other users who wish to contribute don't have to have clang-format installed, and if regular contributors forget to run formatting, it will be caught and updated downstream. This is what the commit message, and author will look like once it has been completed (if any formatting needs to be done)

image

I've taken the "Run Autoformatter" message from previous times when splewis has run the autoformatter on the main branch, so I figured it is a good enough message to show intent.

Will mark ready when 0.10 is released.

### DIFF
--- a/.github/workflows/auto_formatter.yml
+++ b/.github/workflows/auto_formatter.yml
@@ -1,0 +1,28 @@
+name: Auto Format Code On Master Merge
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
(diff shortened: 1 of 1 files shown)


PhlexPlexico:
Decided to update the commit message a tiny bit to give a bit more of a descriptive reason as to why the action would run and commit something to master.

Issue #201, opened 2017-11-14
Coaches aren't included in the ready system


ghost:
So as the title says, coaches aren't included in the ready system and their team tag doesn't show up during the game. If this is not going to be fully implemented, there could be a cvar like in pugsetup to whether include or exclude coaches/spectators from the ready system.

JesusWithPants:
What version ar You using?
Im using the coach mode also on my 5x5 servers, but have no problems with them. Tags appear propperly.

About Ready system - You need to include in the match config: `"min_spectators_to_ready": 0,`, set it to 1 if You have always one coach in server.

slawteruk:
Are you able to confirm if a coach uses a player slot, or a spectator slot? What should a coach be treated as?

nickdnk:
@PhlexPlexico Do we want this? Looking at the code, the new changes to coaches does not seem to consider a coach a "ready player" at all: https://github.com/splewis/get5/blob/master/scripting/get5/readysystem.sp#L85-L103

PhlexPlexico:
I'm not too sure. I'd be interested to see what ESEA does, but my initial thoughts is that they didn't care about coach ready-ups. If that's not the case, I can definitely look into this!

PhlexPlexico:
So I took a look into it and I think ESEA does require the coach to be ready as well. I think get5 is a bit more functional in regards to this as well, so I think we should include coaches in the ready system. Should be something as simple as just removing those `!IsClientCoaching(i)` calls? Will test this and see if it works as intended.

nickdnk:
It looks like we could just remove that coaching filter, yes.

But a thing to consider here: The ready system has the "minimum players ready" (I forget the exact name and I'm on my phone) option. If this is set to 5, and 4 players and a coach ready, the team will be ready without its 5th player.

If we do this, we should probably add a config parameter for it, I.e. `g_CoachesMustReady` and if that is true, override minimum players required for ready and always set it to the length of the player array + length of coach array. How this would work with in scrim-mode I don't really know.

Thoughts?

nickdnk:
We could also just put in the docs that the "minimum players to ready" set to 0 means all players on the team, including coaches.

nickdnk:
This is in preview on the nightly builds for 0.11 now: https://github.com/splewis/get5/releases/tag/v0.11.0-74ed6a8

PR #861, opened 2022-09-02
Add coaches to ready system via optional parameter

nickdnk:
This fixes https://github.com/splewis/get5/issues/201.

~I also got rid of the "format clan tags every single second no matter the game state" code that was running.~

### DIFF
--- a/scripting/get5/matchconfig.sp
+++ b/scripting/get5/matchconfig.sp
@@ -10,6 +10,7 @@
 #define CONFIG_SPECTATORSNAME_DEFAULT "casters"
 #define CONFIG_NUM_MAPSDEFAULT 3
 #define CONFIG_SKIPVETO_DEFAULT false
+#define CONFIG_COACHES_MUST_READY_DEFAULT false
 #define CONFIG_CLINCH_SERIES_DEFAULT true
 #define CONFIG_VETOFIRST_DEFAULT "team1"
 #define CONFIG_SIDETYPE_DEFAULT "standard"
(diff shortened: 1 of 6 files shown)


scripting/get5/matchconfig.sp
```diff
@@ -175,14 +175,24 @@ bool LoadMatchConfig(const char[] config, bool restoreBackup = false) {
         }
       }
     }
+    CreateTimer(1.1, Timer_SetClanTagsAfterMatchLoad, _, TIMER_FLAG_NO_MAPCHANGE);
```
PhlexPlexico, line 178: Any specific reason for setting this to 1.1 seconds?

nickdnk: The timer placing players is 1 second. See above.

PhlexPlexico: Thanks GitHub for shrinking the diff to avoid the giant comment block that would help make more sense of this :) okay perfect! LGTM!

PhlexPlexico (approved):

nickdnk:
So I've thought about this, and I think trying to figure out **all** possible places where a clan tag update could be needed is a lot more complicated than just letting the function run every second, even though I really don't like that approach much either - but it seems to work fine. I suppose having it constantly check also makes sure that people cannot change their tags during a match.

I'm considering rolling back the change that changes this, and just sticking to the coach inclusion part, which was what I originally set out to fix anyway.

PR #862, opened 2022-09-05
Get rid of g_WaitingForRoundBackup and introduce backup game state + clean up associated map-change logic and timers

nickdnk:
I never liked the combination of game being in livemode with `g_WaitingForRoundBackup`. I think it makes more sense to have a gamestate that indicates that we're not live, but pending a restore to a live round.

This changes that. Untested so far.

It's a breaking change because of the new game state, but other than that everything should work the same, the code is just a little less messy in my opinion.


nickdnk:
Works fine. Also cleaned up some timer logic a bit to prevent some potential bugs.

nickdnk:
@splewis You cool with this?

PR #868, opened 2022-09-09
Implemented bearer token parameter to get5_loadmatch_url

Apfelwurm:
To increase security possibilities, we have implemented the ability to set a bearer token to authenticate when using get5_loadmatch_url.
Done by https://github.com/TheR00st3r and me.

### DIFF
--- a/documentation/docs/commands.md
+++ b/documentation/docs/commands.md
@@ -94,9 +94,11 @@ time a backup file is written. Empty string if no backup was written.
 :   Loads a [team section of a match configuration](../match_schema) from a file into a team relative to the `csgo`
 directory. The file must contain a `Get5MatchTeam` object.
 
-####`get5_loadmatch_url <url>` {: #get5_loadmatch_url }
+####`get5_loadmatch_url <url> [headername] [headervalue]` {: #get5_loadmatch_url }
 :   Loads a remote (JSON-formatted) [match configuration](../match_schema) by sending an HTTP(S) `GET` to the given URL.
-You should put the `url` argument inside quotation marks (`""`).
+You should put the `url` argument inside quotation marks (`""`). You may optionally provide an HTTP header and value
+pair using the `headername` and `headervalue` arguments. For example, if you wanted to authenticate your request with a
+Bearer token, you could do: `get5_loadmatch_url "https://example.com" "Authorization" "Bearer <token>"`.
 
 !!! warning "SteamWorks required"
 
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -1114,15 +1114,32 @@ static Action Command_LoadMatchUrl(int client, int args) {
     ReplyToCommand(client,
                    "Cannot load matches from a url without the SteamWorks extension running");
   } else {
-    char arg[PLATFORM_MAX_PATH];
-    if (args >= 1 && GetCmdArgString(arg, sizeof(arg))) {
-      if (!LoadMatchFromUrl(arg)) {
+    char url[PLATFORM_MAX_PATH];
+    if (args >= 1 && GetCmdArg(1, url, sizeof(url))) {
+      ArrayList headerNames;
+      ArrayList headerValues;
+      if (args == 2) {
+        ReplyToCommand(client, "This command accepts either one or three parameters.");
+        return;
+      }
+      if (args == 3) {
+        char headerbuffer[PLATFORM_MAX_PATH];
+        GetCmdArg(2, headerbuffer, sizeof(headerbuffer));
+        headerNames = new ArrayList(PLATFORM_MAX_PATH); 
+        headerNames.PushString(headerbuffer);
+        GetCmdArg(3, headerbuffer, sizeof(headerbuffer));
+        headerValues = new ArrayList(PLATFORM_MAX_PATH); 
+        headerValues.PushString(headerbuffer);
+      }
+      if (!LoadMatchFromUrl(url, _, _, headerNames, headerValues)) {
         ReplyToCommand(client, "Failed to load match config.");
       } else {
         ReplyToCommand(client, "Match config loading initialized.");
       }
+      delete headerNames;
+      delete headerValues;
     } else {
-      ReplyToCommand(client, "Usage: get5_loadmatch_url <url>");
+      ReplyToCommand(client, "Usage: get5_loadmatch_url <url> [headername] [headervalue]");
     }
   }
 }
--- a/scripting/get5/matchconfig.sp
+++ b/scripting/get5/matchconfig.sp
@@ -259,7 +259,8 @@ static void MatchConfigFail(const char[] reason, any...) {
 }
 
 stock bool LoadMatchFromUrl(const char[] url, ArrayList paramNames = null,
-                            ArrayList paramValues = null) {
+                            ArrayList paramValues = null, ArrayList headerNames = null,
+                            ArrayList headerValues = null) {
   bool steamWorksAvaliable = LibraryExists("SteamWorks");
 
   char cleanedUrl[1024];
@@ -279,9 +280,26 @@ stock bool LoadMatchFromUrl(const char[] url, ArrayList paramNames = null,
       return false;
     }
 
+    if (headerNames != null && headerValues != null) {
+      if (headerNames.Length != headerValues.Length) {
+        MatchConfigFail("request headerNames and headerValues size mismatch");
+        delete request;
+        return false;
+      }
+
+      char headerparam[128];
+      char headervalue[1024];
+      for (int i = 0; i < headerNames.Length; i++) {
+        headerNames.GetString(i, headerparam, sizeof(headerparam));
+        headerValues.GetString(i, headervalue, sizeof(headervalue));
+        SteamWorks_SetHTTPRequestHeaderValue(request, headerparam, headervalue);
+      }
+    }
+
     if (paramNames != null && paramValues != null) {
       if (paramNames.Length != paramValues.Length) {
         MatchConfigFail("request paramNames and paramValues size mismatch");
+        delete request;
         return false;
       }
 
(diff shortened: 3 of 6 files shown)


nickdnk:
This is a good idea, but I would prefer we made it more generic, so that you would provide the header as a key-value pair, i.e.

`get5_loadmatch_url "Authorization" "Bearer <tokenhere>"`

This way, people could use whatever type of authorization they want in any format.

scripting/get5/matchconfig.sp
```diff
@@ -277,6 +277,16 @@ stock bool LoadMatchFromUrl(const char[] url, ArrayList paramNames = null,
       return false;
     }
 
+    LogDebug("token (SteamWorks) = %s", token);
+
+    if(strlen(token) > 0){
+      char cleanedToken[1024];
+      strcopy(cleanedToken, sizeof(cleanedToken), token);
+      Format(cleanedToken, sizeof(cleanedToken), "Bearer %s", cleanedToken);
```
nickdnk, line 285: This `cleanedToken` part is also meaningless. Nothing is being cleaned; it's just copied. You could achieve the same with just:

```c
char formattedToken[1024];
Format(formattedToken, sizeof(formattedToken), "Bearer %s", token);
```

scripting/include/get5.inc
```diff
@@ -81,7 +81,7 @@ native bool Get5_LoadMatchConfig(const char[] filename);
 // Loads a match config by downloading a remote config via a HTTP GET request.
 // Request parameters should be left out of the url and put into the optional ArrayLists parameters.
 native bool Get5_LoadMatchConfigFromURL(const char[] url, ArrayList paramNames = null,
-                                        ArrayList paramValues = null);
+                                        ArrayList paramValues = null, const char[] token = "");
```
nickdnk, line 84: The parameter provided here is never read by `Native_LoadMatchConfigFromURL`. Please see how the arraylists are read using `GetNativeCell(x)` - the same should be applied to the header; and preferably a header key-value pair as two arguments, as I commented originally. The current implementation leaves this argument always unused.

Apfelwurm: I should not do stuff without enough sleep.
I should not do stuff without enough sleep.
I should not do stuff without enough sleep.
:D

scripting/get5/matchconfig.sp
```diff
@@ -257,7 +257,8 @@ static void MatchConfigFail(const char[] reason, any...) {
 }
 
 stock bool LoadMatchFromUrl(const char[] url, ArrayList paramNames = null,
-                            ArrayList paramValues = null) {
+                            ArrayList paramValues = null, ArrayList headerNames = null,
+                            ArrayList headerValues = null) {
```
nickdnk, line 261: The header name and value should not be arraylists, as there is always only one string in each. It should be `char[]` buffers.

nickdnk: Unless you intentionally want these to be able to accept multiple values.

Apfelwurm: I've done that intentionally, because for me it made sense to implement this with the most flexibility in mind. 
Also since the params are also implemented like that but they are not used from the command line function, the headers can also be used in the same way if you use the natives from somewhere. Or do you see any problem with this implementation?

scripting/get5/matchconfig.sp
```diff
@@ -277,6 +278,21 @@ stock bool LoadMatchFromUrl(const char[] url, ArrayList paramNames = null,
       return false;
     }
 
+    if (headerNames != null && headerValues != null) {
+      if (headerNames.Length != headerValues.Length) {
+        MatchConfigFail("request headerNames and headerValues size mismatch");
+        return false;
```
nickdnk, line 284: `delete request`

scripting/get5/matchconfig.sp
```diff
@@ -277,6 +278,21 @@ stock bool LoadMatchFromUrl(const char[] url, ArrayList paramNames = null,
...
+      for (int i = 0; i < headerNames.Length; i++) {
+        headerNames.GetString(i, headerparam, sizeof(headerparam));
+        headerValues.GetString(i, headervalue, sizeof(headervalue));
+        SteamWorks_SetHTTPRequestHeaderValue(request, headerparam, headervalue);
+      }      
+    }
+    
     if (paramNames != null && paramValues != null) {
       if (paramNames.Length != paramValues.Length) {
         MatchConfigFail("request paramNames and paramValues size mismatch");
```
nickdnk, line 298: `delete request`

scripting/get5/natives.sp
```diff
@@ -126,7 +126,9 @@ public int Native_LoadMatchConfigFromURL(Handle plugin, int numParams) {
   GetNativeString(1, url, sizeof(url));
   ArrayList paramNames = view_as<ArrayList>(GetNativeCell(2));
   ArrayList paramValues = view_as<ArrayList>(GetNativeCell(3));
-  return LoadMatchFromUrl(url, paramNames, paramValues);
+  ArrayList headerNames = view_as<ArrayList>(GetNativeCell(4));
+  ArrayList headerValues = view_as<ArrayList>(GetNativeCell(5));
+  return LoadMatchFromUrl(url, paramNames, paramValues, headerNames, headerValues);
```
nickdnk, line 131: Since `LoadMatchFromUrl` does not `delete` any of the parameters passed to it, we must:

`delete paramNames;`
`delete paramValues;`
`delete headerNames;`
`delete headerValues;`

after calling `LoadMatchFromUrl`.

nickdnk (changes requested):
Okay, so this is alright, but we need to clean up some memory. See the comments.

Also, in `SteamWorks_OnMatchConfigReceived`, we must `delete` the request handle when returning.

I know these are not your fault, but since we're modifying this function, its memory leaks should be resolved.

scripting/get5/natives.sp
nickdnk: No, wait, sorry. This is wrong. Whoever calls `Get5_LoadMatchConfigFromURL` should do this.

nickdnk:
Hello @Apfelwurm 

You must rebase your branch on develop before I can review it.

Apfelwurm:
@nickdnk done :)

scripting/get5.sp
```diff
@@ -1089,9 +1089,23 @@ static Action Command_LoadMatchUrl(int client, int args) {
     ReplyToCommand(client,
                    "Cannot load matches from a url without the SteamWorks extension running");
   } else {
-    char arg[PLATFORM_MAX_PATH];
-    if (args >= 1 && GetCmdArgString(arg, sizeof(arg))) {
-      if (!LoadMatchFromUrl(arg)) {
+    char url[PLATFORM_MAX_PATH];
+    if (args >= 1 && GetCmdArg(1, url, sizeof(url))) {
+      ArrayList headerNames = new ArrayList(PLATFORM_MAX_PATH);
```
nickdnk, line 1094: These array lists can be declared here (`ArrayList headerNames;`), but you don't need to assign to them (`new`) unless they are needed below, and you must also `delete` both of these after you call `LoadMatchFromURL` below.

nickdnk (changes requested):

Apfelwurm:
@nickdnk done

scripting/get5.sp
```diff
@@ -1089,13 +1089,28 @@ static Action Command_LoadMatchUrl(int client, int args) {
...
+      ArrayList headerNames;
+      ArrayList headerValues;
+      if (args == 2) {
+        ReplyToCommand(client, "This command accepts either one or three parameters.");
+        return;
+      }
+      if (args == 3) {
+        char headerbuffer[PLATFORM_MAX_PATH];
+        GetCmdArg(2, headerbuffer, sizeof(headerbuffer));
+        headerNames.PushString(headerbuffer);
```
nickdnk, line 1103: `headerNames` and `headerValues` are now `null`, as you declared them but did not create the lists (`new ArrayList...`). Move creation of both lists inside the `if (args == 3)`-check.

Apfelwurm: @nickdnk done

documentation/docs/commands.md
```diff
@@ -94,10 +94,13 @@ time a backup file is written. Empty string if no backup was written.
 :   Loads a [team section of a match configuration](../match_schema) from a file into a team relative to the `csgo`
 directory. The file must contain a `Get5MatchTeam` object.
 
-####`get5_loadmatch_url <url>` {: #get5_loadmatch_url }
+####`get5_loadmatch_url <url> [headername] [headervalue]` {: #get5_loadmatch_url }
```
nickdnk, line 97: Change this whole section to this:

```
####`get5_loadmatch_url <url> [headername] [headervalue]` {: #get5_loadmatch_url }
:   Loads a remote (JSON-formatted) [match configuration](../match_schema) by sending an HTTP(S) `GET` to the given URL.
You should put the `url` argument inside quotation marks (`""`). You may optionally provide an HTTP header and value
pair using the `headername` and `headervalue` arguments. For example, if you wanted to authenticate your request with a
Bearer token, you could do: `get5_loadmatch_url "https://example.com" "Authorization" "Bearer <token>"`.
```

scripting/get5.sp
```diff
@@ -1114,13 +1114,30 @@ static Action Command_LoadMatchUrl(int client, int args) {
...
+      }
+      if (!LoadMatchFromUrl(url, _, _, headerNames, headerValues)) {
         ReplyToCommand(client, "Failed to load match config.");
       } else {
         ReplyToCommand(client, "Match config loading initialized.");
       }
+      delete headerNames;
+      delete headerValues;
     } else {
       ReplyToCommand(client, "Usage: get5_loadmatch_url <url>");
```
nickdnk, line 1142: Final thing. Add optional headers in response here.

nickdnk (changes requested):

PR #869, opened 2022-09-09
0.10.1

nickdnk:
As https://github.com/splewis/get5/pull/867 introduces some breaking changes, I took out the essential parts that fix broken stuff only.

This still has the fixes for warmup during back instead of a restart + not kicking secondary GOTV, but it does not make any changes to the behavior of `get5_check_auths`, and instead fixes the `!swap` command when auths are not enforced.

### DIFF
--- a/documentation/docs/getting_started.md
+++ b/documentation/docs/getting_started.md
@@ -7,12 +7,11 @@
 
 ## Quick Start {: #quick-start }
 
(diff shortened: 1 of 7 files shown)


nickdnk:
This breaks the coaching and team-name assignment logic quite a bit. I would still prefer the approach where we lock the auths when loading a match.

Issue #873, opened 2022-09-12
Cannot enter get5_web_api_url because double forward slash starts comment


paoloposo:
I cannot figure out how to set the `get5_web_api_url` cvar. If I enable debug messages with `get5_debug 1` and then try to set the url with `get5_web_api_url "http://somehost"`, I get this in the console:

    [get5_api] get5_web_api_url now set to https:/

Everything after the colon is cut (and then get5 adds the trailing slash). [This](https://developer.valvesoftware.com/wiki/CFG) states that everything after two forward slashes is regarded as a comment. **So how can I enter the URL then?** Am I missing something blatantly obvious? I searched for how to escape forward slashes, but couldn't find anything. I also tried escaping them with a backward slash, but no luck.

Plugin version is 0.10.1.

PhlexPlexico:
So your example is a bit confusing here - I assume in your actual implementation you're using https? Where are you entering this value? In a config, or setting via console? I believe if you wrap it in quotes it should work. As an example using rcon

```
rcon get5_web_api_url https://somehost.com/
```

Should set it on the server. You can then pipe it back by just running `rcon get5_web_available`

If you're using a config, using quotes should be enough to avoid cfg files treating it as a comment.

Apfelwurm:
This is strange. Do you plan to load a config in the future? That way it is working for sure currently

nickdnk:
Can you try it without `rcon`, directly on the server, just to rule out that this is causing problems?

paoloposo:
Hm, when I tried it using the server conf before it didn't work. Must have done something wrong because it works now that I tried it again. I made a config file with the following content

    get5_web_api_url "https://somehost/"

and then executed using `rcon exec`. That worked. Now I get this output:

```
] rcon exec set_api_url.cfg
[get5_api] get5_web_api_url now set to https://somehost/
] rcon get5_web_api_url
"get5_web_api_url" = "https://somehost" ( def. "" )            - URL the get5 api is hosted at
] rcon get5_web_available
{
    "available": 1,
    "gamestate": 7,
    "plugin_version": "0.10.1"
}
```

Instead of somehost I have actually used a host where a web server is listening. But it doesn't get any requests. The game server is running in a container (using the `cm2network/csgo` image) and I verified that the web server can be reached by running a curl from within the container. Do I have to configure any more settings for the stats to be sent? I didn't yet start any game via the get5 configuration files. I just used the "Create match with current players" option from the chat menu.

[...]

PR #875, opened 2022-09-12
fix request memory leak and check for http status codes + added debug logging to apistats when no api key was provided

Apfelwurm:
when no api key is setted in get5_apistats, CreateRequest just returns INVALID_HANDLE and nothing happens, so i added debug logging in there. Related to: https://github.com/splewis/get5/issues/873#issuecomment-1244574027

Also since i touched apistats, i tried to implement the changes @nickdnk already suggested in #868

Please review my changes carefully, i might have deleted a request in a wrong place or smth like that, since i'm not able to test the changes at the moment.

### DIFF
--- a/scripting/get5/natives.sp
+++ b/scripting/get5/natives.sp
@@ -21,6 +21,7 @@ public APLRes AskPluginLoad2(Handle myself, bool late, char[] error, int err_max
   CreateNative("Get5_AddLiveCvar", Native_AddLiveCvar);
   CreateNative("Get5_IncreasePlayerStat", Native_IncreasePlayerStat);
   CreateNative("Get5_GetMatchStats", Native_GetMatchStats);
+  CreateNative("Get5_CreateGet5HTTPRequest", Native_CreateGet5HTTPRequest);
   RegPluginLibrary("get5");
   return APLRes_Success;
 }
@@ -250,4 +251,11 @@ public int Native_GetMatchStats(Handle plugin, int numParams) {
     g_StatsKv.Rewind();
     return view_as<int>(true);
   }
+}
+
+public int Native_CreateGet5HTTPRequest(Handle plugin, int numParams) {
+  EHTTPMethod method = view_as<EHTTPMethod>(GetNativeCell(1));
+  char url[1024];
+  GetNativeString(2, url, sizeof(url));
+  return view_as<int>(CreateGet5HTTPRequest(method, url));
 }
--- a/scripting/get5_apistats.sp
+++ b/scripting/get5_apistats.sp
@@ -112,21 +112,20 @@ void ApiInfoChanged(ConVar convar, const char[] oldValue, const char[] newValue)
 }
 
 static Handle CreateRequest(EHTTPMethod httpMethod, const char[] apiMethod, any:...) {
-  char url[1024];
-  FormatEx(url, sizeof(url), "%s%s", g_APIURL, apiMethod);
 
-  char formattedUrl[1024];
-  VFormat(formattedUrl, sizeof(formattedUrl), url, 3);
-
-  LogDebug("Trying to create request to url %s", formattedUrl);
-
-  Handle req = SteamWorks_CreateHTTPRequest(httpMethod, formattedUrl);
   if (StrEqual(g_APIKey, "")) {
     // Not using a web interface.
+    LogError("Failed to create request because get5_web_api_key was not provided");
     return INVALID_HANDLE;
+  }
 
-  } else if (req == INVALID_HANDLE) {
-    LogError("Failed to create request to %s", formattedUrl);
+  char url[1024];
+  char formattedUrl[1024];
+  FormatEx(url, sizeof(url), "%s%s", g_APIURL, apiMethod);
+  VFormat(formattedUrl, sizeof(formattedUrl), url, 3);
+  LogDebug("Trying to create request to url %s", formattedUrl);
+  Handle req = Get5_CreateGet5HTTPRequest(httpMethod, formattedUrl);
+  if (req == INVALID_HANDLE) {
     return INVALID_HANDLE;
 
   } else {
@@ -136,15 +135,28 @@ static Handle CreateRequest(EHTTPMethod httpMethod, const char[] apiMethod, any:
   }
 }
 
-int RequestCallback(Handle request, bool failure, bool requestSuccessful,
+static int RequestCallback(Handle request, bool failure, bool requestSuccessful,
                     EHTTPStatusCode statusCode) {
+
   if (failure || !requestSuccessful) {
-    LogError("API request failed, HTTP status code = %d", statusCode);
-    char response[1024];
-    SteamWorks_GetHTTPResponseBodyData(request, response, sizeof(response));
-    LogError(response);
+    LogError("Network connection failed for the API request");
+    delete request;
     return;
   }
+
+  int status = view_as<int>(statusCode);
+  if (status >= 300 || status < 200) {
+    LogError("API request failed with HTTP status code: %d.", statusCode);
+    int responseSize;
+    SteamWorks_GetHTTPResponseBodySize(request, responseSize);
+    char[] response = new char[responseSize];
+    if (SteamWorks_GetHTTPResponseBodyData(request, response, responseSize)) {
+      LogError("Response body: %s", response);
+    } else {
+      LogError("Failed to read response body.");
+    }
+  }
+  delete request;
 }
 
 public void Get5_OnSeriesInit(const Get5SeriesStartedEvent event) {
@@ -196,17 +208,31 @@ static void CheckForLogo(const char[] logo) {
   }
 }
 
-static int LogoCallback(Handle request, bool failure, bool successful, EHTTPStatusCode status,
-                        int data) {
+static int LogoCallback(Handle request, bool failure, bool successful, EHTTPStatusCode statusCode,
+                        DataPack data) {
   if (failure || !successful) {
-    LogError("Logo request failed, status code = %d", status);
+    LogError("Network connection failed for the logo request");
+    delete request;
     return;
   }
 
-  DataPack pack = view_as<DataPack>(data);
-  pack.Reset();
+  int status = view_as<int>(statusCode);
+  if (status >= 300 || status < 200) {
+    LogError("Logo request failed with HTTP status code: %d.", statusCode);
+    int responseSize;
+    SteamWorks_GetHTTPResponseBodySize(request, responseSize);
+    char[] response = new char[responseSize];
+    if (SteamWorks_GetHTTPResponseBodyData(request, response, responseSize)) {
+      LogError("Response body: %s", response);
+    } else {
+      LogError("Failed to read response body.");
+    }
+    delete request;
+  }
+
+  data.Reset();
   char logo[32];
-  pack.ReadString(logo, sizeof(logo));
+  data.ReadString(logo, sizeof(logo));
 
   char logoPath[PLATFORM_MAX_PATH + 1];
   if (g_UseSVGCvar.BoolValue) {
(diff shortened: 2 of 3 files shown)


scripting/get5_apistats.sp
```diff
@@ -123,10 +123,13 @@ static Handle CreateRequest(EHTTPMethod httpMethod, const char[] apiMethod, any:
   Handle req = SteamWorks_CreateHTTPRequest(httpMethod, formattedUrl);
   if (StrEqual(g_APIKey, "")) {
     // Not using a web interface.
+    LogDebug("Failed to create request because no ApiKey was provided");
+    delete req;
     return INVALID_HANDLE;
 
   } else if (req == INVALID_HANDLE) {
     LogError("Failed to create request to %s", formattedUrl);
+    delete req;
```
nickdnk, line 132: You don't need to delete something that's already an `INVALID_HANDLE`.

scripting/get5_apistats.sp
```diff
@@ -123,10 +123,13 @@ static Handle CreateRequest(EHTTPMethod httpMethod, const char[] apiMethod, any:
   Handle req = SteamWorks_CreateHTTPRequest(httpMethod, formattedUrl);
   if (StrEqual(g_APIKey, "")) {
     // Not using a web interface.
+    LogDebug("Failed to create request because no ApiKey was provided");
```
nickdnk, line 126: Make this a `LogError` instead and make it more descriptive of what is actually missing. Put in the name of the Cvar so people know what to adjust.

scripting/get5_apistats.sp
```diff
@@ -138,11 +141,18 @@ static Handle CreateRequest(EHTTPMethod httpMethod, const char[] apiMethod, any:
 
 int RequestCallback(Handle request, bool failure, bool requestSuccessful,
                     EHTTPStatusCode statusCode) {
-  if (failure || !requestSuccessful) {
-    LogError("API request failed, HTTP status code = %d", statusCode);
-    char response[1024];
-    SteamWorks_GetHTTPResponseBodyData(request, response, sizeof(response));
-    LogError(response);
+  int status = view_as<int>(statusCode);
+  if (failure || !requestSuccessful || status >= 300 || status < 200) {
```
nickdnk, line 145: If `failure` or `!requestSuccesful`, you should not check for HTTP status code or attempt to read the body. This needs to be broken into two parts.

scripting/get5_apistats.sp
```diff
@@ -184,6 +194,7 @@ static void CheckForLogo(const char[] logo) {
                      : CreateRequest(k_EHTTPMethodGET, "/static/img/logos/%s.png", logo);
 
     if (req == INVALID_HANDLE) {
+      delete req;
```
nickdnk, line 197: Unnecessary delete. Put in a LogDebug to indicate the request failed to start.

scripting/get5_apistats.sp
```diff
@@ -196,10 +207,20 @@ static void CheckForLogo(const char[] logo) {
   }
 }
 
-static int LogoCallback(Handle request, bool failure, bool successful, EHTTPStatusCode status,
+static int LogoCallback(Handle request, bool failure, bool requestSuccessful, EHTTPStatusCode statusCode,
                         int data) {
-  if (failure || !successful) {
-    LogError("Logo request failed, status code = %d", status);
+  int status = view_as<int>(statusCode);
```
nickdnk, line 212: Same as above. Break into two and don't read response or HTTP code if unsuccessful request.

scripting/get5_apistats.sp
```diff
@@ -231,6 +256,7 @@ public void Get5_OnGoingLive(const Get5GoingLiveEvent event) {
     AddStringParam(req, "mapname", mapName);
     SteamWorks_SendHTTPRequest(req);
   }
+  delete req;
```
nickdnk, line 259: Don't delete here. Delete in the callback.

scripting/get5_apistats.sp
```diff
@@ -246,6 +272,7 @@ static void UpdateRoundStats(const char[] matchId, const int mapNumber) {
     AddIntParam(req, "team2score", t2score);
     SteamWorks_SendHTTPRequest(req);
   }
+  delete req;
```
nickdnk, line 275: Delete in callback.

scripting/get5_apistats.sp
```diff
@@ -279,6 +306,7 @@ public void Get5_OnMapResult(const Get5MapResultEvent event) {
     AddStringParam(req, "winner", winnerString);
     SteamWorks_SendHTTPRequest(req);
   }
+  delete req;
```
nickdnk, line 309: Delete in callback.

scripting/get5_apistats.sp
```diff
@@ -337,6 +365,7 @@ static void UpdatePlayerStats(const char[] matchId, const int mapNumber, const K
         AddIntStat(req, kv, STAT_MVP);
         SteamWorks_SendHTTPRequest(req);
       }
+      delete req;
```
nickdnk, line 368: Delete in callback.

scripting/get5_apistats.sp
```diff
@@ -375,6 +404,7 @@ public void Get5_OnSeriesResult(const Get5SeriesResultEvent event) {
     AddIntParam(req, "forfeit", forfeit);
     SteamWorks_SendHTTPRequest(req);
   }
+  delete req;
```
nickdnk, line 407: Delete in callback.

scripting/get5_apistats.sp
```diff
@@ -138,11 +141,18 @@ static Handle CreateRequest(EHTTPMethod httpMethod, const char[] apiMethod, any:
...
+    LogError("Steamworks GET request failed with HTTP status code: %d.", statusCode);
+    int responseSize;
+    SteamWorks_GetHTTPResponseBodySize(request, responseSize);
+    char[] response = new char[responseSize];
+    if (SteamWorks_GetHTTPResponseBodyData(request, response, responseSize)) {
+      LogError("Response body: %s", response);
+    } else {
+      LogError("Failed to read response body.");
+    }
+    delete request;
```
nickdnk, line 155: Unnecessary `return`, and you should delete the request also if it has failed, so it must go outside your error check.

nickdnk (changes requested):

nickdnk:
Related: https://github.com/splewis/get5/issues/873

nickdnk:
@Apfelwurm Could you finish this one?

Apfelwurm:
@nickdnk tbh i`m currently too busy because of an upcoming event that we host next week, so i can get to it after the 10th october.

scripting/get5_apistats.sp
```diff
@@ -123,6 +123,8 @@ static Handle CreateRequest(EHTTPMethod httpMethod, const char[] apiMethod, any:
   Handle req = SteamWorks_CreateHTTPRequest(httpMethod, formattedUrl);
   if (StrEqual(g_APIKey, "")) {
     // Not using a web interface.
+    LogError("Failed to create request because get5_web_api_key was not provided");
```
nickdnk, line 126: Use the method in `http.sp` to create request handles. And move the API string check before creating the handle.

scripting/get5_apistats.sp
```diff
@@ -196,13 +212,28 @@ static void CheckForLogo(const char[] logo) {
   }
 }
 
-static int LogoCallback(Handle request, bool failure, bool successful, EHTTPStatusCode status,
+static int LogoCallback(Handle request, bool failure, bool successful, EHTTPStatusCode statusCode,
```
nickdnk, line 215: You can change the data argument to `DataPack` instead of `int`, that way you won't have to cast it below.

scripting/get5_apistats.sp
```diff
@@ -196,13 +212,28 @@ static void CheckForLogo(const char[] logo) {
...
                         int data) {
   if (failure || !successful) {
-    LogError("Logo request failed, status code = %d", status);
+    LogError("Logo request failed, status code = %d", statusCode);
```
nickdnk, line 218: Status code is always 0 here, as it won't have received an HTTP response at all. Change it to something like "Network connection failed." or something.

scripting/get5_apistats.sp
```diff
@@ -184,6 +199,7 @@ static void CheckForLogo(const char[] logo) {
                      : CreateRequest(k_EHTTPMethodGET, "/static/img/logos/%s.png", logo);
 
     if (req == INVALID_HANDLE) {
+      LogDebug("Failed start the logo API request");
```
nickdnk, line 202: Redundant LogDebug as the built-in `http.sp` method already logs if it returns an invalid handle.

scripting/get5_apistats.sp
```diff
@@ -138,13 +140,26 @@ static Handle CreateRequest(EHTTPMethod httpMethod, const char[] apiMethod, any:
 
 int RequestCallback(Handle request, bool failure, bool requestSuccessful,
```
nickdnk, line 141: You can make this `static`

nickdnk (changes requested):

scripting/get5_apistats.sp
Apfelwurm: done

scripting/get5_apistats.sp
Apfelwurm: done

scripting/get5_apistats.sp
Apfelwurm: done

scripting/get5_apistats.sp
Apfelwurm: done

scripting/get5_apistats.sp
Apfelwurm: done

Apfelwurm:
@nickdnk you can take a look on everything, should be okay now but i have to test it tomorrow

scripting/get5/natives.sp
```diff
@@ -251,3 +252,10 @@ public int Native_GetMatchStats(Handle plugin, int numParams) {
     return view_as<int>(true);
   }
 }
+
+public int Native_CreateGet5HTTPRequest(Handle plugin, int numParams) {
```
nickdnk, line 256: We need to add a check here for the SteamWorks library, but we can't run it all the time as it's expensive, so we have to have a boolean/int check. Is there a specific reason this returns an `int` and not a `Handle?` Must a Native return an integer? If yes, just cast `INVALID_HANDLE` as int here.

```c
static int hasSteamWorks = 0;
if (hasSteamWorks != 1 && (hasSteamWorks == -1 || !LibraryExists("SteamWorks"))) {
  LogError("The SteamWorks extension is required in order to make HTTP requests with Get5.");
  hasSteamWorks = -1;
  return INVALID_HANDLE;
}
hasSteamWorks = 1;
```

scripting/include/get5.inc
```diff
@@ -5,6 +5,7 @@
 
 #include <json>  // github.com/clugg/sm-json
 #include <cstrike>
+#include <SteamWorks>
```
nickdnk, line 8: With this include, we should probably remove the include in `get5.sp`, as that file already includes `get5.inc`. Not sure about this though.

scripting/get5_apistats.sp
```diff
@@ -196,17 +208,31 @@ static void CheckForLogo(const char[] logo) {
   }
 }
 
-static int LogoCallback(Handle request, bool failure, bool successful, EHTTPStatusCode status,
-                        int data) {
+static int LogoCallback(Handle request, bool failure, bool successful, EHTTPStatusCode statusCode,
+                        DataPack data) {
   if (failure || !successful) {
-    LogError("Logo request failed, status code = %d", status);
+    LogError("Network connection failed for the logo request");
```
nickdnk, line 214: Missing full stop. I would also recommend you move the `DataPack` logic up to the beginning of this method, so you can error out which request failed, i.e:

```c
LogError("Network connection failed for the request for logo at '%s'.", logo);
```

You are leaking memory now if the request fails, as you are not deleting the datapack at all then.

nickdnk (changes requested):

nickdnk:
@Apfelwurm Should we just discard this? It's so far behind on commits now that's it's probably almost a do-over anyway.

Apfelwurm:
> @Apfelwurm Should we just discard this? It's so far behind on commits now that's it's probably almost a do-over anyway. The apistats extension is also not supported or developed on at all.

Most likeley that would make sense. Kind of sad that i did not finish it quick enough, but i`m super low on time in the last months.

PR #876, opened 2022-09-14
0.10.2

nickdnk:
It seems there are some problems related to spectators and the priority of team-selection if a player is defined in multiple places in a match configuration.

1. This change ensures that team assignment is always in this priority:

    1. Spectator
    2. Coach for team1
    3. Coach for team2
    4. Player on team1
    5. Player on team2

2. Spectators cannot ".coach" and they will be removed if they exceed the `mp_spectators_max` cvar.
3. Renamed `GetTeamAuths` to `GetTeamPlayers`, so it has more consistent naming, as the one for coaches simply is `GetTeamCoaches`. Renamed `g_TeamAuths` to `g_TeamPlayers` as well.
4. `GetAuthMatchCoachTeam` has been merged into `GetMatchTeamFromAuth` and renamed for a single-point-of-entry that handles the logical flow for which team you belong to.
5. Consistently reset the player auth arrays when loading teams or spectator dictionaries. This was a bit hit-and-miss before.
6. Allow spectators in scrim templates, which this new logic allows.
7. Avoid duplicate calls to `GetClientMatchTeam` in `GetPlayerStat` + `GetPlayerStat` when incrementing stats. Tested and works.

I would like to release this as 0.10.2.

### DIFF
--- a/scripting/get5/teamlogic.sp
+++ b/scripting/get5/teamlogic.sp
@@ -17,6 +17,15 @@ void CheckClientTeam(int client) {
     return;
   }
 
+  if (correctTeam == Get5Team_Spec) {
+    if (CountPlayersOnTeam(correctTeam, client) >= FindConVar("mp_spectators_max").IntValue) {
+      KickClient(client, "%t", "TeamIsFullInfoMessage");
+    } else {
+      SwitchPlayerTeam(client, Get5Side_Spec);
+    }
+    return;
+  }
+
   Get5Side correctSide = view_as<Get5Side>(Get5TeamToCSTeam(correctTeam));
   if (correctSide == Get5Side_None) {
     // This should not be possible.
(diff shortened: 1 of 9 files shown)


scripting/get5/teamlogic.sp
```diff
@@ -279,33 +283,31 @@ Get5Team CSTeamToGet5Team(int csTeam) {
   }
 }
 
-Get5Team GetAuthMatchTeam(const char[] steam64) {
+Get5Team GetAuthMatchTeam(const char[] steam64, bool includeCoaches = true, bool includePlayers = true) {
```
PhlexPlexico, line 286: `GetMatchTeamFromAuth` maybe? This current call makes it sound like you're retrieving the steam ID from the MatchTeam.

PR #877, opened 2022-09-14
Optimize calls to stat increments

nickdnk:
When I was working on https://github.com/splewis/get5/pull/876, I noticed that we do this:

```c
int value = GetPlayerStat(client, field);
return SetPlayerStat(client, field, value + delta);
```

 - which triggers a call to `GetClientMatchTeam` for both reading the value **and** writing it just after. This doubles the calls to this function which has to auth and then check all team arrays to find out where a player belongs. This is already time-consuming enough, so doing it twice for absolutely no reason is not very clever.

### DIFF
--- a/scripting/get5/stats.sp
+++ b/scripting/get5/stats.sp
@@ -921,15 +921,10 @@ static Action Stats_RoundMVPEvent(Event event, const char[] name, bool dontBroad
   }
 }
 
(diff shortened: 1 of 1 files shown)


nickdnk:
Will fix in https://github.com/splewis/get5/pull/876 instead.

PR #878, opened 2022-09-15
Fix Portuguese translation

SidiBecker:
Hello! Here is some changes to portuguese phrases!

### DIFF
--- a/translations/pt/get5.phrases.txt
+++ b/translations/pt/get5.phrases.txt
@@ -6,11 +6,11 @@
     }
     "WaitingForCastersReadyInfoMessage"
     {
-        "pt"            "Esperando pelos streammers para digitar {2} pra começar."
+        "pt"            "Esperando pelos streamers digitar {2} pra começar."
     }
     "ReadyToRestoreBackupInfoMessage"
     {
-        "pt"            "Digite {1} quando vocês estiver pronto para restaurar o backup da partida."
+        "pt"            "Digite {1} quando vocês estiverem prontos para restaurar o backup da partida."
     }
     "ReadyToKnifeInfoMessage"
     {
@@ -54,15 +54,15 @@
     }
     "MinutesToForfeitMessage"
     {
-        "pt"            "{1} tem {2} minutos para ficar prontos ou eles se renderão."
+        "pt"            "{1} tem {2} minutos para ficarem prontos ou eles se renderão."
     }
     "SecondsToForfeitInfoMessage"
     {
-        "pt"            "{1} tem {2} segundos para ficar prontos ou eles se renderão."
+        "pt"            "{1} tem {2} segundos para ficarem prontos ou eles se renderão."
     }
     "10SecondsToForfeitInfoMessage"
     {
-        "pt"            "{1} tem 10 segundos para fica prontos ou se render."
+        "pt"            "{1} tem 10 segundos para ficarem prontos ou se render."
     }
     "MaxPausesUsedInfoMessage"
     {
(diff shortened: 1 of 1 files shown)


nickdnk:
Thanks @SidiBecker 

Can you add the ones missing also? If you compare the Portuguese translation to the English, you'll see that it's incomplete. You can find a full list of translation strings for the development version here: https://splewis.github.io/get5/dev/translations/#reference

If you don't want to, that's fine and we can just accept this also.

SidiBecker:
> Thanks @SidiBecker
> 
> Can you add the ones missing also? If you compare the Portuguese translation to the English, you'll see that it's incomplete. You can find a full list of translation strings for the development version here: https://splewis.github.io/get5/dev/translations/#reference
> 
> If you don't want to, that's fine and we can just accept this also.

Sure! I'll do it.

nickdnk:
Great. Starting from the bottom of that list will probably give you the most missing strings to begin with, but you'll have to go through them all as I'm not sure where there are holes.

SidiBecker:
@nickdnk I just added the missing phrases.

translations/pt/get5.phrases.txt
```diff
@@ -6,11 +6,11 @@
     }
     "WaitingForCastersReadyInfoMessage"
     {
-        "pt"            "Esperando pelos streammers para digitar {2} pra começar."
+        "pt"            "Esperando pelos streamers digitar {2} pra começar."
```
nickdnk, line 9: Are you sure the right word is "streamers" here? It's the spectators/commentators, also called "casters", that must be ready. @SidiBecker

nickdnk:
`NextSeriesMapInfoMessage` is missing its "seconds" argument. This was added later, so that's probably why you missed it.

translations/pt/get5.phrases.txt
SidiBecker: Yeah, that's wrong. It should be use {1} param there. I'll fix.

nickdnk: Also that.

nickdnk: No, sorry. It's `{2}` because the first argument is the name of the team, i.e. "Casters", for the casters. But since you defined the casters simply as "streamers", without the name they were given in the match, the argument is not used.

SidiBecker: `"Aguardando {1} digitar {2} pra começar."` would be right?

nickdnk: I would assume, yes, with the reservation that I don't speak Portuguese.

translations/pt/get5.phrases.txt
```diff
@@ -248,4 +356,84 @@
     {
         "pt"            "[NOT READY]"
```
nickdnk, line 357: PRONT and NÃO PRONTO perhaps? Or would you not say that?

translations/pt/get5.phrases.txt
SidiBecker: So yes, its like `"Waiting for {1} to type {2} to begin."`. I'll fix.

translations/pt/get5.phrases.txt
SidiBecker: Ready is a well-known term, I wouldn't change it

SidiBecker: Also NÃO PRONTO is strange to say

SidiBecker:
Updated!

nickdnk:
You can try this development version out with this translation: https://github.com/splewis/get5/releases/tag/v0.11.0-0c5c721

SidiBecker:
Thanks!

nickdnk:
Note that development versions may be buggy. It is not recommended that you keep using this, so it's just for testing that the translation works like you want it. It will be released with the official version 0.11 when that's done.

PR #881, opened 2022-09-16
Use get5_server_id to prevent backup file collisions

nickdnk:
If you run a containerized system, you may have collisions in the built-in backup files the game produces on its own.

This sets the `mp_backup_round_file` to `backup_%d`, instead of the default `backup`, so that `get5_server_id` can be used to suffix an integer to these files. A file called `backup_round00.txt` would then be `backup_4_round00.txt` instead.

Also, the original backup files are now deleted after Get5 copies them into its own backup file.

### DIFF
--- a/documentation/docs/backup.md
+++ b/documentation/docs/backup.md
@@ -31,13 +31,13 @@ They print in the format `filepath timestamp team1 team2 map team1_score team2_s
 
 ```
 > get5_listbackups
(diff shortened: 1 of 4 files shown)


nickdnk:
Tested. Confirmed working.

nickdnk:
Maybe we should add `get5_server_id` to the backup file format also, i.e. `get5_backup%d_match%s_map%d_round%d.cfg`

splewis:
Looks good to me.

splewis (approved):

PR #883, opened 2022-09-18
Allow for custom chat aliases

nickdnk:
This allows users to set their own custom chat commands without modifying the get5 source files at all. This is done via a KeyValue file currently just placed at `addons/sourcemod/configs/get5/commands.cfg`, but this could also be a ConVar.

Setting a custom command will also override the output in the messages get5 prints, so if you set your `!ready` command to be `!start`, the message would instead be:

`Type !start when you are ready to begin.`

The way it currently works is just that the **last** command that was set determines the output, but they all work, hence this as the default:

```c
MapChatCommand(Get5ChatCommand_Ready, "r");
MapChatCommand(Get5ChatCommand_Ready, "ready");
```

### DIFF
--- a/scripting/get5.sp
+++ b/scripting/get5.sp
@@ -153,6 +153,7 @@ ArrayList g_MapPoolList;
 ArrayList g_TeamPlayers[MATCHTEAM_COUNT];
 ArrayList g_TeamCoaches[MATCHTEAM_COUNT];
 StringMap g_PlayerNames;
+StringMap g_ChatCommands;
 char g_TeamNames[MATCHTEAM_COUNT][MAX_CVAR_LENGTH];
 char g_TeamTags[MATCHTEAM_COUNT][MAX_CVAR_LENGTH];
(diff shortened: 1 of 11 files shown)


scripting/get5.sp
```diff
@@ -711,7 +712,8 @@ static Action Timer_InfoMessages(Handle timer) {
   }
 
   char readyCommandFormatted[64];
-  FormatChatCommand(readyCommandFormatted, sizeof(readyCommandFormatted), "!ready");
+  GetChatAliasForCommand(Get5ChatCommand_Ready, readyCommandFormatted, sizeof(readyCommandFormatted));
+  FormatChatCommand(readyCommandFormatted, sizeof(readyCommandFormatted), readyCommandFormatted);
```
nickdnk, line 716: This is how it would be applied to all commands across the plugin for output in messages, I just only updated this "!ready" example for the PR so far.

PR #885, opened 2022-09-21
Allow sending events to remote URL

nickdnk:
As the current event system is thoroughly documented as JSON, but requires SourceMod knowledge to actually use, as you would have to write your own plugin to hook onto events, I figured it would be a nice addition to Get5 if it could natively send all events somewhere over HTTP. This lets this happen.

No breaking changes, just 3 cvars:

`get5_remote_log_url` to enable logging.

`get5_remote_log_header_key` and `get5_remote_log_header_value` for authorization, if required.

Also moved some common code from the demo upload logic to an `http.sp` file.

Untested.

### DIFF
--- a/documentation/docs/configuration.md
+++ b/documentation/docs/configuration.md
@@ -58,9 +58,10 @@ cfg/get5/live.cfg # (3)
 ## Server Setup
 
 ####`get5_server_id`
-:   Integer that identifies your server. This is used in temporary and backup files to prevent collisions. You should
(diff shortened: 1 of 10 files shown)


nickdnk:
Tested. Works fine.

PhlexPlexico (approved):
Took a good read into all of this. Good changes, nothing to really comment on, so this should be good to go in whenever I think 👍

PR #888, opened 2022-09-28
Fix problems with knife swap/stay command timing

nickdnk:
It seems there are still some rare problems with `!stay` and `!swap` being called at **the exact wrong time** causing the game to misbehave and sometimes skip the live countdown entirely.

This PR disables the ability to choose a side until the knife-round has completely ended, removing the race-condition window that would cause this bug.

It also reduces info message interval from 29 to 20, but instead restarts the timer on each round start to avoid double-printing messages, such as the "Team X won the knife round, waiting for them to type swap/stay", which is also triggered when the knife round ends.

Added `mp_round_restart_delay 3` to the default knife config as you now can't do swap/stay until the round ends, so 5 seconds is a bit much.

Deduplicate logic for the knife-decision chat message

Remove round-end-fun-fact override hack; it was buggy.

I would like to merge this to 0.10 and release as 0.10.4 (with https://github.com/splewis/get5/pull/887), as this is potentially game-breaking and 0.11 is still not ready.

### DIFF
--- a/cfg/get5/knife.cfg
+++ b/cfg/get5/knife.cfg
@@ -10,3 +10,4 @@ mp_roundtime 1.92
 mp_roundtime_defuse 1.92
 mp_roundtime_hostage 1.92
 mp_t_default_secondary ""
+mp_round_restart_delay 3
--- a/documentation/docs/configuration.md
+++ b/documentation/docs/configuration.md
@@ -117,7 +117,8 @@ in [scrim mode](../getting_started/#scrims), put on `team2`. **`Default: 1`**
 
 ####`get5_time_to_make_knife_decision`
 :   Time (in seconds) a team has to make a [`!stay`](../commands/#stay) or [`!swap`](../commands/#swap)
-decision after winning knife round, 0 = unlimited. **`Default: 60`**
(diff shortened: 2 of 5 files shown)


nickdnk:
Also, I have tested this as well as I can with just myself and a bot, so I know it works, but I would appreciate it if someone else would give it a test.

nickdnk:
I observed that doing knife cvar restore when the knife round reaches its win-panel is the cause of some of these problems. It will potentially restore `mp_respawn_on_death_t/ct` (which is set in warmup, then saved as a cvar overridden by knife config) right as the win panel shows up which causes players to sometimes respawn. I've moved this to the same place as the warmup config which seems to solve that problem. It also makes sense that we don't mess with cvars for a round that has not yet completely ended. This may have been the cause of a bunch of other issues as well, as these cvars could contain basically anything, as configs are user-defined.

nickdnk:
I have extensively tested this in real games and manual testing and everything seems to be working correctly.

splewis (approved):
LGTM; this is a good catch. I'm not surprised such a race condition was possible.

PR #889, opened 2022-09-30
Use SteamWorks_SetHTTPRequestContextValue to pass demo file name

nickdnk:
Instead of `g_DemoFileNameLastStartedUpload`, this allows us to associate the filename with the request without introducing any global state.


nickdnk:
Tested. Works fine.

nickdnk:
Also added `FileExists` check to prevent exception, as apparently it throws if the file does not exist.

PhlexPlexico (approved):

PR #891, opened 2022-10-04
Disallow suicide via console

nickdnk:
This disables access to `kill` and `explode` via console.

1. You should not be suiciding in a real match. It is also disabled in official matchmaking.
2. Suiciding at various times, such as right as the knife round starts, causes problems.


Apfelwurm (approved):

Iwhite67:
I agree but give possibility to disabled

I make all get5 test With kill

nickdnk:
> I agree but give possibility to disabled
> 
> I make all get5 test With kill 

I would not like to make it possible to disable it when it is a potential cause of problems. For testing this also makes no sense, as you would potentially get a false positive on a bug. If you want to end a round quickly, set the round time very low or use `sv_cheats` to noclip and win.

PR #895, opened 2022-10-16
Add more event data

nickdnk:
1. Adds `time_until_restore` to `Get5_OnSeriesResult`, so a server management system knows how long to wait until the server is available after a match ends.
2. Adds `Get5_OnDemoUploadEnded` which fires when the demo upload either fails or succeeds, again letting a server management system know when this has completed so the server can be shut down.

Tested and works as expected.

### DIFF
--- a/documentation/docs/event_schema.yml
+++ b/documentation/docs/event_schema.yml
@@ -261,6 +261,14 @@ paths:
                     SourceMod.
                 winner:
                   $ref: "#/components/schemas/Get5Winner"
+                time_until_restore:
+                  type: integer
+                  minimum: 0
+                  example: 45
+                  description: |
+                    The number of seconds until Get5 restores any changed ConVars after the series has ended. This is
+                    the GOTV broadcast delay plus a few seconds. If you use this event to relinquish a server instance,
+                    you should wait until this time has passed before starting a new match or shutting down the server.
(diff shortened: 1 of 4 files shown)


nickdnk:
@PhlexPlexico Any opinions on this? I was a little conflicted about the naming of some of these things, for instance; I didn't want to call it "Get5_OnDemoUploadCompleted" as it also fires when it fails, and that would imply success.

PhlexPlexico:
LGTM. We could try `Get5_OnDemoUploadFinished` instead? Finishing can either indicate failure or completion as well, and it keeps in line with the other `Get5_OnDemoFinished`? If not, this still looks acceptable :)

nickdnk:
> LGTM. We could try `Get5_OnDemoUploadFinished` instead? Finishing can either indicate failure or completion as well, and it keeps in line with the other `Get5_OnDemoFinished`? If not, this still looks acceptable :)

I did think of that, but I still feel like "finished" implies success **and** it also means it also has the same name as the other one, which might cause them to be confused for one another.

PhlexPlexico:
That's fair then. It's just cutting into the weeds of semantics, so this should be good to merge in :)

PhlexPlexico (approved):

PR #905, opened 2022-10-18
Update french translation

Iwhite67:


translations/fr/get5.phrases.txt
```diff
@@ -106,11 +106,11 @@
     }
     "TimeRemainingBeforeAnyoneCanUnpausePrefix"
     {
-        "fr"            "Temps restant avant de pouvoir mettre fin à la pause "
```
Gryfenfer97: The reason why I have put a space here is because the next character is a colon (cf. https://splewis.github.io/get5/dev/translations/#reference)
In french we have to put a space before and after a colon.
https://fr.m.wikisource.org/wiki/Aide:Guide_typographique

Iwhite67: Correct
The correct is "la pause : "

nickdnk (changes requested):
Looks good to me, given that I don't speak French. Only this minor thing.

PR #906, opened 2022-10-19
Add delay for team placement when joining the server

nickdnk:
It appears that not having this delay is causing problems for some servers, so I have reintroduced it here for testing.
I've tested this on my end, and it works fine with this delay, so if this solves the problem for those that have it and can reproduce it (I cannot), we can merge this in.


nickdnk:
Confirmed to be fixing the issue.

PR #908, opened 2022-10-20
Don't trigger auto-forfeit during warmup/veto + add option to disable forfeit

nickdnk:
The auto-forfeit timer is now never started during warmup/veto. You will now have to use `get5_time_to_start` if you want that phase to be limited in duration instead. Previously, the surrender timer would still start during warmup **if** `get5_time_to_start` was 0, which is both confusing/hard to document and also meant that matches may easily be incorrectly ended if someone joins during warmup and leaves for whatever reason, technical or other, without coming back within the grace period. It was not a good design decision.

I also removed a bit of redundancy in the timer-logic and increased the default forfeit countdown (`get5_forfeit_countdown`) from 60 to 180 seconds.


scripting/get5/surrender.sp
```diff
@@ -255,13 +223,9 @@ void CheckForForfeitOnDisconnect() {
...
   Get5Team forfeitingTeam;
-
-  if (g_GameState <= Get5State_Warmup && g_TeamTimeToStartCvar.IntValue > 0) {
-    // If we're in warmup or veto and a "time to ready" value is defined; let that handle it.
-    return;
-  } else if (team1Count > 0 && team2Count > 0) {
+  if (team1Count > 0 && team2Count > 0) {
     // If both teams still have at least one player; check for single-disconnect tech trigger.
-    if (g_GameState == Get5State_Live && g_AutoTechPauseCvar.IntValue == view_as<int>(Get5AutoTechPauseType_AnyPlayer)) {
+    if (g_AutoTechPauseCvar.IntValue == view_as<int>(Get5AutoTechPauseType_AnyPlayer)) {
```
nickdnk, line 228: Reminder: Get rid of the `else` for this one and move `Get5Team forfeitingTeam;` below it. Readability is low here.

nickdnk:
Reminder: If one team leaves, the other team leaves, first team comes back and their timer is stopped, a new timer if the other team is still empty should start.

PR #910, opened 2022-10-21
Don't consume pauses until they start + GSI support + Pause begin forward/event

nickdnk:
Current logic causes a pause to be consumed for a team immediately as a player types `!pause` or `!tech`. This changes this so the pause isn't consumed until the pause actually starts, and so that it can be `!unpause`'ed before the round ends. I'm not entirely sure we want this, but here it is as a proposal. Thoughts?

I also reintroduced the GSI for fixed tactical pauses, as that *may* be best for some tournaments, so the fixed pause duration can be displayed in a custom overlay. This does mean, however, that the pause timers are doubled for players.


nickdnk:
Reminder to myself: Pausing during knife consumes a pause but does not actually pause. Work is still needed.

LukasW1337:
As someone who uses get5 for tournaments, i support using the GSI when having a defined tactical timeout,, This just makes sense broadcast wise, and for the players in-game.

Don't exactly know how to interpret this:
> This does mean, however, that the pause timers are doubled for players.

Is it meant that it shows the timer both in UI and in chat?

nickdnk:
This has been rebased on dev and is now ready for some testing.

I've introduced a few changes:

1. The default pause logic is now fixed a duration of 60 (`get5_fixed_pause_time 60`) instead of a combined 300 seconds (`get5_max_pause_time 300`), which is more in line with FaceIT, Matchmaking and any pro game. For anyone upgrading on an existing server, this does not matter, as it's just the default values that are changed.
2. You can now set `get5_allow_unpausing_fixed_pauses 0` if you don't want to allow `!unpause` during fixed-duration pauses.
3. The GSI for pauses seems to be working as intended now. I need someone using GOTV overlays with the GSI to actually test if it behaves as expected.
4. If you pause during the knife or post-knife, the Get5 hints will still be used, but they will be switched for the GSI pauses once the game is live. The GSI for pauses does not work during warmup, but the game still pauses, so this is kind of a workaround that seems to check all the boxes.

nickdnk:
After some testing:

1. The GSI now works in all cases I can realistically think of. We had an issue where the timeout was toggled but the game was not actually paused according to gameprops, which confused some GSI overlays to not indicate a pause. This has been resolved.

2. A forward called `Get5_OnPauseBegan` has been added, which fires when a pause actually starts. This _could_ be used to mute certain players during pauses. Closes https://github.com/splewis/get5/issues/924.

3. A message has been added as feedback about `!stop` not being available during when called freezetime. Players are expected to use `!tech` at this stage, and getting confirmation fro the other team before freezetime runs out is also difficult anyway. If the game is paused for backup, `!stop` *can* be used multiple times during freeztime if required.

scripting/get5/pausing.sp
```diff
@@ -169,7 +164,7 @@ Action Command_Pause(int client, int args) {
 
   int maxPauses = g_MaxTacticalPausesCvar.IntValue;
 
-  if (!g_FixedPauseTimeCvar.BoolValue) {
+  if (g_FixedPauseTimeCvar.IntValue < 15) {
```
nickdnk, line 167: Reminder: Revert this.

scripting/get5/pausing.sp
```diff
@@ -349,6 +386,11 @@ static Action Timer_PauseTimeCheck(Handle timer) {
       ConvertSecondsToMinutesAndSeconds(maxTacticalPauseTime, pauseTimeMaxFormatted, sizeof(pauseTimeMaxFormatted));
     }
 
+    if (fixedPauseTime > 0 && g_GameState == Get5State_Live) {
+      // Fixed-duration pauses use the GSI during live.
```
nickdnk, line 390: This should be warmup check.

nickdnk (commented):
Minors.

nickdnk:
Squashed, ready for merge pending some testing.

PR #914, opened 2022-10-23
Add option to postpone stopping the demo recording

nickdnk:
It seems that some servers cannot handle doing GOTV broadcast **and** flushing demos at the same time. This adds an option to extend the delay on flushing demos on round end, allowing for the full match to be broadcast on GOTV before the demo is stopped, even with a long delay.


nickdnk:
It might actually just be better to make this a boolean convar and make it 1:1 to the tv_delay if true. It appears this is a universal problem that happens if you have a substantial tv_delay.

nickdnk:
Converted to boolean. Updated docs.

Issue #509, opened 2020-04-30
get5_hostname_format gets reset on map change and is not persisted for hostname setting


topjor:
### Expected behavior
When a matchconfig is loaded and has a cvar for `get5_hostname_format` or if this variable is changed (via console or rcon) this is persisted trough loading a match with mapchange event before the match loading is finished.

### Actual behavior
When the matchconfig is loaded, the map gets changed before setting the server hostname. while mapchanging is happening `get5_hostname_format` gets reset to its default value and the hostname gets set with the default format and not the altered format.

### Steps to reproduce
- Plugin version: 0.7.1
- Sourcemod version: 1.10

Steps to reproduce:
1. Change `get5_hostname_format` by console/rcon or using a matchconfig cvar
2. Load a matchconfig with a `skip_veto` enabled and a different map then the server has loaded currently
3. read out the current server hostname (I used the console `hostname` command)

(tested with `get5_loadmatch_url`)
I have only tested BO1, I have no idea about the re-setting on mapchange while in multi-map matches.
The mapchange on matchconfig load only happens when skip_veto is set and another map set to be played.

Issue #438, opened 2019-10-22
Hostname not being restored after match finish


mrc4tt:
Hi,

It could be nice to get little function on get5 plugin, with a little change after the match is finished, so "hostname" aka servername will be changed back as normal from "server.cfg"

Example:
Match1 > Hostname: Get5: Mikkel vs Test1 
After 16-0 - where the match is finished.
Server is still having "Get5: Mikkel vs Test1" as servername.
The server should have changed the server name to what it has from server.cfg:
Example: 
1) Get5: Mikkel vs Test1 (Match)
2) Mikkel's Server (After Match is finished) - exec from the server.cfg (from get5 plugin)

splewis:
Are you sure that's not *already* happening? get5 has some code already that saves previous cvar values before execution & restores them later, but things get a bit confusing since there's both the warmup & live config. 

Can you make sure to check the actual ``hostname`` cvar in console? The display on the scoreboard isn't an accurate reflection of the cvar, the last time I checked (it only seemed to update on map loads).

MikkelDK:
It doesn't change the hostname of Get5: Mikkel vs Test1 to Mikkel's Server (from server.cfg) after the match is finished. which it doesn't do.

Aka:
```
  g_SetHostnameCvar = CreateConVar(
      "get5_hostname_format", "Get5: {TEAM1} vs {TEAM2}",
```
**After MATCH is done with 16-0** - **It's still running with Get5: Test1 vs Test2** - My question is: Isn't it supposed to change the server name back as it was called when the game is over, or what?

**Example:**
**Game1: Hostname will be: Get5: spelwis vs mikkel**
**Game1 is finish 16-0 to spelwis** - Here is the problem, **It don't exec server.cfg OR change hostname back as it was before match was started!**

splewis:
> After MATCH is done with 16-0 - It's still running with Get5: Test1 vs Test2

I understand that. I'm asking how you're determining what "it" is. Are you:
1) Looking at the scoreboard in-game
2) Checking the value of the cvar via console (like `sm_cvar hostname` or such)

If you think there is indeed a bug, you should post steps to reproduce as well.

MikkelDK:
Checking the value of the cvar via console (like sm_cvar hostname or such) & checking the server on favorites (via Steam)

DennisF10:
I can confirm that hostname doesn't change back, no matter if you check server browser, scoreboard or sm_cvar hostname etc.

ndstudios-admin:
[...]

PR #933, opened 2022-11-16
Fix hostname issues + add score tags

nickdnk:
This fixes 5 issues:

1. Fix https://github.com/splewis/get5/issues/509
2. Fix https://github.com/splewis/get5/issues/438
3. `mp_teamscore_max` is now being properly reset if going from Bo3/5 to Bo1.
4. You can now include `{TEAM1_SCORE}` and `{TEAM2_SCORE}` in your hostname (or anywhere else these tags can be used), and hostname will update on each round start.
5. Spaces in team names are no longer replaced by underscore in hostname.

### DIFF
--- a/documentation/docs/configuration.md
+++ b/documentation/docs/configuration.md
@@ -313,9 +313,9 @@ exist.<br>**`Default: ""`**
 disable.<br>**`Default: "get5_matchstats_{MATCHID}.cfg"`**
 
 ####`get5_hostname_format`
-:   The hostname to apply to the server when a match configuration is loaded.
-[State substitutes](#state-substitutes) can be used. Leave blank to disable changing the hostname.<br>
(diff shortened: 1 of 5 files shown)


misc/format.py
```diff
@@ -23,7 +23,7 @@ def main():
                 files.append(os.path.join(dirpath, f))
 
     for filename in files:
-        subprocess.call('clang-format --style=file:{0}/misc/.clang-format -i {1}'.format(root_path, filename), shell=True)
+        subprocess.call('clang-format --style=file:./misc/.clang-format -i {0}'.format(filename), shell=True)
```
PhlexPlexico, line 26: Does this fix the relative path issues now with the workflow?

PhlexPlexico (approved):

misc/format.py
nickdnk: No. That still does not work.

Issue #604, opened 2020-12-18
Managing map data from separate files


deLuXe83:
Hi,
for team data, it's possible to use the "fromfile" field to includes external files.

Would it be possible to use this command accordingly for the maplist section??

Thanks!!

PR #946, opened 2022-11-28
Refactor loading match configs + tests

nickdnk:
Similarly to teams, this allows the maplist to load using the `fromfile` logic.

Closes https://github.com/splewis/get5/issues/604.


nickdnk:
So, this does a few other things:

1. I added **a lot** of unit tests for loading match configurations and checking for various errors.
2. I use the `SMCParser` to validate KeyValue files before importing them, as `KeyValues.ImportFromFile` returns true even if the file is invalid (which makse no sense, but it's the game engine doing this, not SM).
3. You can now use `fromfile` for both `map_list`, `spectators` and `team1`/`team2`, all supporting both the KeyValue format and JSON.
4. The error feedback when loading match configurations with errors in them should be much better at pointing at the exact error.
5. The bug with the captain not being at index 0 when loading JSON teams has been fixed.
6. Logic that handles `mp_team_xx` variables has been cleaned and the documentation has been updated. This means you can now actually set `matchtext` for a team in a BoX and not have it overwritten by their team map series score.
7. ConVars can now be up to 512 characters instead of 128. Note that the source may be a limiting factor, as you can load ConVars with a longer value from a plugins .cfg file than you can from the console input.
8. When using `clinch_series: false`, you can now have up to the full maplist of points in the GOTV map score indicator, i.e. bo5 without clinch grants you 5 point slots intead of 3, which would be the max with clinch.
9. All Get5 state variables are now reset when a series ends. This should reduce the risk of game-state related bugs.

nickdnk:
Also fixes a bug that would sometimes cause players to be put on the wrong team if restoring from the first round of the second half to any round in the first half.

scripting/get5/matchconfig.sp
```diff
@@ -457,24 +426,19 @@ static void AddTeamBackupData(KeyValues kv, Get5Team team) {
...
   g_SkipVeto = kv.GetNum("skip_veto", CONFIG_SKIPVETO_DEFAULT) != 0;
   g_CoachesMustReady = kv.GetNum("coaches_must_ready", CONFIG_COACHES_MUST_READY_DEFAULT) != 0;
-
   g_NumberOfMapsInSeries = kv.GetNum("num_maps", CONFIG_NUM_MAPSDEFAULT);
-  g_MapsToWin = MapsToWin(g_NumberOfMapsInSeries);
-  if (g_NumberOfMapsInSeries != 2 && g_NumberOfMapsInSeries % 2 == 0) {
-    MatchConfigFail("Cannot create a series of %d maps. Use an odd number or 2.", g_NumberOfMapsInSeries);
-    return false;
-  }
+  g_MapsToWin = g_SeriesCanClinch ? MapsToWin(g_NumberOfMapsInSeries) : g_NumberOfMapsInSeries;
```
PhlexPlexico, line 441: Do we no longer fail on best of series that do not fit the format of BO3/5/7?

PhlexPlexico (approved):
Just this one little comment of the Best Of Series allowing a BO4 to happen (from what I can read from the logic), but other than that everything looks good to me. Good job on the tests, they feel very comprehensive and I don't think there's anything I can really add to them!

scripting/get5/matchconfig.sp
nickdnk: Well, no, because we have the `series_clinch` parameter now, which can be disabled if you just want to play any number of maps, including any size of odd numbers. We *could* fail it if you have series clinch enabled and don't play an odd number, but I kind of feel like this is up to the user to decide if they want or not.

PhlexPlexico: Sounds good to me then :)

Issue #953, opened 2022-12-12
[Feature Request] Add configuration for wingman


Cyberschorsch:
It would be great to have a default config for running wingman matches.

Since some users here and in Discord were asking for it I would like to start this discussion here. I will also open up a Pull Request with a config I use for competive wingman matches we use in our internal cup.

nickdnk:
As commented on the PR:

This is a little more complicated than this, unfortunately. Game mode and game type must be correctly set, and the location of the wingman config file needs to be a Cvar like the others. I was planning on properly integrating this in 0.13, after 0.12 is done.

PR #954, opened 2022-12-12
feat(#953): Add wingman configuration

Cyberschorsch:
This pull request adds a new cfg for running wingman matches.


Cyberschorsch:
This will resolve #953

nickdnk:
This is a little more complicated than this, unfortunately. Game mode and game type must be correctly set, and the location of the wingman config file needs to be a Cvar like the others. I was planning on properly integrating this in 0.13, after 0.12 is done.

PR #964, opened 2023-01-13
fix(include): correct syntax errors

hammy2899:
Fix syntax error preventing compiling on latest sourcemod version

### DIFF
--- a/scripting/include/get5.inc
+++ b/scripting/include/get5.inc
@@ -1729,7 +1729,7 @@ stock void ConvertBombSiteToStringInJson(const JSON_Object obj, const char[] key
 }
 
 // Called each get5-event with JSON formatted event text.
-forward void Get5_OnEvent(const Get5Event, const char[] eventJson);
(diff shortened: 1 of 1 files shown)


nickdnk:
Hello

If you want to fix this, you have to make all required changes for SM 1.11 throughout the plugin, not just this one line. It's on the to-do for version 0.13. or 0.14. Until then, compile with 1.10.

scripting/include/get5.inc
```diff
@@ -1729,7 +1729,7 @@ stock void ConvertBombSiteToStringInJson(const JSON_Object obj, const char[] key
 }
 
 // Called each get5-event with JSON formatted event text.
-forward void Get5_OnEvent(const Get5Event, const char[] eventJson);
+forward void Get5_OnEvent(Get5Event event, const char[] eventJson);
```
nickdnk, line 1732: Add `const`, i.e. `forward void Get5_OnEvent(const Get5Event event, const char[] eventJson);`

hammy2899:
@nickdnk should be good now

Issue #755, opened 2022-05-30
[Feature Request] Attempt to download a workshop map


GekasD:
When a map is being loaded, it should check if the map name is a workshop id (e.x 1729981703) and then run `host_workshop_map <mapid>` instead of `changelevel`, so then it would be possible to pass in workshop map ids in the maplist of a match config.

GekasD:
https://github.com/splewis/get5/blob/master/scripting/get5/maps.sp

Im guessing the check would need to happen on the 2nd function in the file linked above.

PhlexPlexico:
Looks like this is already done in pugsetup - 

https://github.com/splewis/csgo-pug-setup/blob/d03a71cd0251731a13c115aadd19e94def6209db/scripting/pugsetup/maps.sp#L35-L39

So I think this is pretty transferable to this project as well!

GekasD:
Yep looks like exactly what i want, hope this gets implemented soon!

PhlexPlexico:
So yeah, per discord discussions having to implement this may be a bit difficult as well though since we want to avoid as many bugs as possible. A few things that would have to be considered:

- Ensure the server has an API auth key set. Not sure how we can do that but having a check would be better then countless troubleshooting posts.
- Network hiccups which causes a map to not load. There would have to be a way to call a retry if the map hasn't changed. Maybe incorporate a timer function call that checks after X seconds and retry if it hasn't changed?

I think those were the main issues posted in the discord.

nickdnk:
Is there a function that returns the current workshop ID of a map? If not, there is no way to implement retry-logic with a timer.

PhlexPlexico:
I don't have the time to test, but maybe something like [`GetCurrentMap`](https://www.sourcemod.net/new-api/halflife/GetCurrentMap) may work and return a workshop ID instead of the map name?

nickdnk:
No, that returns the name of the map.

GekasD:
What do you mean by current workshop id of a map?, also did you try GetCurrentMap on a workshop map? it returns it in the format workshopid/mapname but i could be wrong

PhlexPlexico:
> What do you mean by current workshop id of a map?, also did you try `GetCurrentMap` on a workshop map? it returns it in the format `workshopid/mapname` or `workshop/workshopid/mapname` but i could be wrong

[...]

PR #974, opened 2023-02-04
Allow workshop maps in maplist

nickdnk:
Closes https://github.com/splewis/get5/issues/755.

Maps can be added to the `maplist` using the syntax `"workshop/map_id/map_name"`, i.e. `"workshop/1193875520/Old Aztec"`.

The name is used in the map selection phase to identify the map.

~One thing left to do here is to put the game into warmup mode when the map ends, if the next map is a workshop map, as it may take 5-10 seconds to download the map.~ Edit: Instead we add 20 seconds to the restart delay, which should be plenty to download a map.


nickdnk:
Tested, working.

PR #978, opened 2023-02-09
Add support for Wingman + clean up default configs + fix mapchange with workshop map

nickdnk:
Closes https://github.com/splewis/get5/issues/953.

I also cleaned up the default configs. I looked at some of the values from https://github.com/splewis/get5/pull/954 - namely the overtime parameters, and merged this with the default `gamemode_competitive2v2.cfg`. I also sorted the default configs and made sure they contain the exact same parameters, but of course with different values between 5v5 and wingman.

A new Cvar is introduced; `get5_live_wingman_cfg`, which points to `get5/live_wingman.cfg` by default (you must copy in this file if you want to test this PR!).

In this draft, you simply set `wingman: true`, `players_per_team: 2` and a Wingman-friendly `maplist`. Get5 will reload the map if the `game_mode` or `game_type` is incorrect, making it trivial to switch between game modes on one server.

I also refactored some of a the map code and removed the constraint from current 0.13 prerelease that disallows duplicate map names - it's no longer necessary to enforce.


nickdnk:
Note to self: Test that backups work as expected with this.

nickdnk:
Also fixes a problem from https://github.com/splewis/get5/pull/974 that would incorrectly format the map name buffer causing map change to simply not work.

PR #980, opened 2023-02-14
Fix double increment of `rounds_played` on first round

nickdnk:
Because the game restarts when going live, the `rounds_played` stat is being incremented one too many.


nickdnk:
Seems to be no bug here anyway.

PR #988, opened 2023-02-28
Upgrade the `!get5` menu and `get5_creatematch`

nickdnk:
This significantly improves the `!get5` menu. It now supports:

1. Changing the game mode (Wingman vs Competitive)
2. Setting the number of players per team (1-7)
3. Setting the number of maps to play
4. Setting the map selection mode
    a. Use current (Bo1 only)
    b. Manual selection
    c. Pick/Ban (Requires Bo2 or more)
5. Setting the team selection mode
    a. Use current teams (requires full teams)
    b. Pick both teams from team file
    c. Scrim; pick only the home team from team file
6. Setting the team captains (requires "use current teams")
7. Friendly Fire on/off
8. Overtime on/off
9. Play all rounds on/off ("practice mode", i.e. play all 30 rounds, not first to 16)
10. Setting the side selection
   a. Standard (Requires pick/ban map selection)
   b. Always knife
   c. Team 1 always CT
   
For team and map selection, you must define your map pool and teams in new config files. This also makes the "scrim mode" a lot more flexible, as you can store all the teams on your server and start a scrim with any of them. Before, you would only be able to scrim with one home team, using the fixed `scrim_template.cfg` file.

Because we now load teams and maps in multiple places (match configurations and the menu), I rewrote the team, map and match config validation logic to be a little more strict and better at giving feedback about wrong parameters.

Update:
I added options to end the match and select a winner (`get5_endmatch team1|team2`).
I added options that lets you browse latest backup files and select the appropriate map/round from the menu.
Updated documentation a lot for all of the above.


nickdnk:
This now also provides a CLI-like structure to `get5_creatematch`, which allows this command to load almost any kind of match configuration.

PR #1000, opened 2023-03-15
Clean up always-reload logic and make it automatic

nickdnk:
Followup to https://github.com/splewis/get5/pull/999, but with option to disable this behavior.

Some more testing is required to determine what the actual problem is here.


nickdnk:
It's better to simply let this happen automatically, which is now achieved by setting a boolean on match_end and map_start. Tested and working as expected; we'll see if it fixes the bug.

PR #1022, opened 2023-04-25
Fix stats for coaches

nickdnk:
When a coach was encountered, the JSON stats would stop filling the player array.


nickdnk:
Confirmed working.

PR #1031, opened 2023-07-01
Fix FCVAR flags and bump 0.15.0

nickdnk:
Fixes some inconsistencies with FCVAR flags.

Minor bump because it's technically a breaking change, although fairly easy to update to with no consequences for most uses.


PhlexPlexico (approved):

PhlexPlexico:
LGTM 👍

Issue #1034, opened 2023-08-01
[Feature Request]: Increase get5_demo_upload_url max length


FlowingSPDG:
### Prereleases & Documentation

- [X] I have looked in [the documentation](https://splewis.github.io/get5) and I don't see the feature anywhere.
- [X] I have looked for [pre-releases](https://github.com/splewis/get5/releases) and the feature has not been added.

### Get5 Version

v0.15.0

### The Feature

Currently I am building cloud based management system that manages srcds and get5 matches.
I am planning to upload demo through Google Cloud Run or something, but it has body length limit(32MiB) for HTTP/1.
So I am going to let srcds upload demo cloud storage directly, but url length would be about 900 chars.
is it possible to increase them?

Thank you for developing and maintaining great plugin for CSGO community!

FlowingSPDG:
Here's match schema for `get5_loadmatch_url` .
https://got5-match-loader-op4haeqloa-an.a.run.app/get5/match/329523134

nickdnk:
As far as I can tell it's just a matter of increasing this constant: https://github.com/splewis/get5/blob/development/scripting/get5/util.sp#L6

I think going to 1025 would probably be fine. I can change this later today for 0.15.1.

FlowingSPDG:
I made PR for it, let me know if we need additional modification!
#1035

PR #1035, opened 2023-08-02
Increase MAX_CVAR_LENGTH(1025)(#1034)

FlowingSPDG:
Increased MAX_CVAR_LENGTH to 1024(+1 buffer).


nickdnk:
I need to go over the places this constant is being used and maybe modify some of them. In the meantime, you can test if the build solves your problem by downloading it from here: https://github.com/splewis/get5/actions/runs/5734168261?pr=1035

The documentation also needs to be updated.

FlowingSPDG:
Got it, I'll take a look more and fix code/docs if I can.
Thanks!

FlowingSPDG:
> Have you checked if this works as expected for you?

Sorry not yet, I'll let you know once I check into it.

PR #1036, opened 2023-08-04
Fix "penetrated" field type on document

FlowingSPDG:
Fix "penetrated" field type on player_death event.


nickdnk:
Hello

`penetrated` is the number of objects the bullet went through. If you want to evaluate it as boolean, just check for `> 0`.

nickdnk:
Sorry, I thought it was changed the other way around; from number to boolean. This is correct.

nickdnk:
Please change it from `number` to `integer` and I will merge it.

nickdnk:
Fixed with https://github.com/splewis/get5/pull/1037 as there was another instance of integer/number error on bomb timer.

